PR #1, opened 2013-02-22
Dev alex

alexcouper:
@txels 

This gets the basic Flask app up and running to be extended.

Potentially `lib` should be renamed `core`. Not sure what you think about that.

Next up for me will be to create some defined interfaces of features to track.


alexcouper:
This now has some defined interfaces for providers. Let me know if you think this is a bad way of doing it.

In theory adding a new provider will be as "easy" as providing a module somewhere on the python path and configuring deploystream to include it (that part isn't actually defined yet in the above PR).

txels:
Ha, Python notoriously lacks a good way to define and enforce interfaces, with its culture of duck-typing.
Have you looked at Zope interfaces? I would like to give that a try.
http://apidoc.zope.org/++apidoc++/Book/ifaceschema/interface/show.html

Take a look and tell me what you think.

alexcouper:
I'm actually quite a fan of the duck-typing way and I'm not over the moon at the prospect of provider-writers needing to import an interface class from deploystream.

But on the other hand it would be very nice for provider-writers to know at the point of writing whether their plugins are going to match the current API.

How would you feel if during the initialization of a plugin it was deploystream's responsibility to check that everything it needed was there - including appropriate args - and failing if they weren't?

Or are there other reasons for needing an interface definition other than: a) helping plugin writers write code and b) ensuring from a main app point of view that we can communicate with the plugins.

alexcouper:
Although my solution doesn't give the developer access to the definition (in code form) at the time of his/her writing. Hmm.....

txels:
Besides how we enforce interfaces, what about using classes rather than
module functions? One base class per type of provider.

This will allow us later to add general methods to base classes etc.

On Fri, Feb 22, 2013 at 4:56 PM, Alex Couper notifications@github.comwrote:

> Although actually my solution doesn't give the developer access at the
> time of writing. Hmm.....
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/pretenders/deploystream/pull/1#issuecomment-13955369.

## 

Carles Barrobés
Lead Developer

 Queens Yard
179a Tottenham Court Road
London W1T 7PA

www.rockabox.com | @rockabox https://twitter.com/Rockabox

alexcouper:
That was my first thought.

I decided against it for the same reason: that it's quite annoying as a third party to need to import a base class from some other project in order to write something for it.

In order to achieve the same result without the annoyance, I foresee these plugin modules being gobbled up by different provider classes that themselves help enforce the interface and we can extend to add methods etc.

txels:
So you must find all frameworks ultra annoying ;) Django, Flask...

On Fri, Feb 22, 2013 at 5:40 PM, Alex Couper notifications@github.comwrote:

> That was my first thought.
> 
> I decided against it for the same reason: that it's quite annoying as a
> third party to need to import a base class from some other project in order
> to write something for it.
> 
> In order to achieve the same result without the annoyance, I foresee these
> plugin modules being gobbled up by different provider classes that
> themselves help enforce the interface and we can extend to add methods etc.
> 
> —
> Reply to this email directly or view it on GitHubhttps://github.com/pretenders/deploystream/pull/1#issuecomment-13957575.

## 

Carles Barrobés
Lead Developer

 Queens Yard
179a Tottenham Court Road
London W1T 7PA

www.rockabox.com | @rockabox https://twitter.com/Rockabox

alexcouper:
I guess so...

To me it feels nicer (for some def of nice) to be able to say to a developer: "No dependencies needed to write your plugin, we just need to be able to call these 2 functions from you. And we will call this one if it's available, but if not no worries".

But maybe that's just me.

alexcouper:
@txels: What are the reasons for requiring a base class dependency when writing a plugin? (as opposed to reading in this module into a base class with methods)
- Quick feedback on missing methods for the developer (possibly)
- ?

deploystream/providers/source_code_control.py
```diff
@@ -0,0 +1,24 @@
+from deploystream.providers.base import ProviderPluginReader
+
+
+class SourceCodeControlPlugin(ProviderPluginReader):
```
alexcouper, line 4: An alternative API for this would be for each base plugin to define the methods it wants functions for and the ProviderPluginReader could use those to check that the arguments match up. We'd still need lists of required and optional functions but it might be clearer that way.

alexcouper:
I shall leave this for you to have a think about and I'm interested to hear what you think. I'm not going to be able to do anything on this until Monday evening at the earliest so I'll leave it in your hands...

deploystream/providers/source_code_control.py
txels, line 4: This is precisely what Zope interfaces can be used for ;)

You can programatically verify implementation of interfaces even if the users don't explicitly declare them:
https://gist.github.com/txels/5019775

This may give us the best of both worlds, our users need not explicitly declare interfaces (although they are free to use them as a tool to check for compliance). Internally we can use them to formally document and verify plugins.

alexcouper, line 4: D'oh. I misunderstood. I thought the implementing class had to state the
interface(s) it was implementing.

OK cool, I'll change the plugin code to use these style interfaces on
Monday.

In the meantime it might make sense to merge this so you can start building
some front end stuff?

In deploystream/providers/source_code_control.py:

> @@ -0,0 +1,24 @@
> +from deploystream.providers.base import ProviderPluginReader
> +
> +
> +class SourceCodeControlPlugin(ProviderPluginReader):

This is precisely what Zope interfaces can be used for ;)

You can programatically verify implementation of interfaces even if the
users don't explicitly declare them:
https://gist.github.com/txels/5019775

This may give us the best of both worlds, our users need not explicitly
declare interfaces (although they are free to use them as a tool to check
for compliance). Internally we can use them to formally document and verify
plugins.

—
Reply to this email directly or view it on
GitHubhttps://github.com/pretenders/deploystream/pull/1/files#r3126682.

txels, line 4: No worries, I did not know either, had to play a little with it ;)

deploystream/providers/examples/template_ci.py
```diff
@@ -0,0 +1,24 @@
+def get_build_information(repo, branch, commit, **kwargs):
```
txels, line 1: I think we want to decouple _conventions_ from the various plugins.
Here we need to pass job name, I think. We also need commit ID to ensure we are checking that the build info matches the correct commit. But not repo and branch.

We need a `conventions` module that will use various rules to figure out branch names, job names, etc. based on regexes and whatelses.

txels, line 1: Or maybe it makes sense to leave that to the plugin. Up for discussion.

deploystream/providers/examples/template_management.py
```diff
@@ -0,0 +1,22 @@
```
txels, line 0: `planning` instead of `management`?

deploystream/providers/examples/template_source_code_control.py
```diff
@@ -0,0 +1,41 @@
+def get_branches_involved(repo_name, feature_id, **kwargs):
```
txels, line 1: similar as comment on CI, maybe here we need to pass regexes that match branch names?
Since the branches will not be created by the plugin, the plugin cannot enforce conventions.
Unless we include a client library with scripts for story management from the command line... the the tool can enforce conventions.

deploystream/providers/examples/template_source_code_control.py
```diff
@@ -0,0 +1,41 @@
...
+        The name of the repository to search for branches.
+
+    :param hierarchy_tree:
+        A tree-like object.
+
+        TODO:
+            more definition here... (traversing, etc)
+
+    """
+    pass
```
txels, line 41: We may want to distinguish between version control (git, mercurial, svn...) and repo hosting (github, bitbucket...).
Even if we implement only git and github for now, the functionality each provides shouldn't be muddled up.

txels:
Still open points for discussion, but happy to merge. Although maybe we should define a branching scheme for this repo? E.g. use a `develop` branch to merge pre-releases, etc...

deploystream/providers/examples/template_ci.py
alexcouper, line 1: I guess I figured that it would be the responsibility of the plugin to figure out what the job name should be given the repo, branch and commit.

Reason being that if you have travis building your PRs then there is no jobname - the appropriate plugin would need to query github's api for the status of the given commit and return that.

But if we want to enforce conventions then in the case where it's querying a single jenkins job it does make sense for this to take some extra conventions fields. But perhaps that can be encapsulated in **kwargs and added to the configuration of the appropriate plugin.

deploystream/providers/examples/template_management.py
alexcouper, line 0: nice. I'll change this...

deploystream/providers/examples/template_source_code_control.py
alexcouper, line 1: I had thought of this function as quite low-level - the only "agreed" convention would be that `feature_id` is somewhere in the branch name. It would be up to the `get_feature_info` in lib.py to sort through the branches (using some regexes in config) into the hierarchy.

Or do you think we can't assume that feature_id will be in the branch names and that some other more complex rules might need to exist at this level?

deploystream/providers/examples/template_source_code_control.py
alexcouper, line 41: We would have a git plugin - to get the merged status we only use a local repo anyway. 

This function will probably need changing into a lower level function like get_merged_status(repo_name, from_branch, to_branch). And called from a higher level one that takes the hierarchy tree and adds the statuses to it.

I didn't envisage github or bitbucket being involved in this call...

alexcouper:
The develop branch seems like a good idea.

PR #5, opened 2013-02-25
Dev alex

alexcouper:
@txels Another big one for you ;)
# What does this add?
- The zope interfaces as requested ;)
- Adds an initial stub of a github plugin
- Adds an extendable end-to-end test with some dummy plugins.
- Adds the core functionality of creating the branch info.
# What still needs to be done?
- Code that handles the creation of hierarchical branch trees
- git plugin code to look across a group of repos.
- git plugin code to use hierarchical trees to determine merge statuses
- Code to handle multiple plugins - we'll need to merge data between them I suspect.
- Code that passes per-plugin settings to the plugins.
- There's inconsistency in the naming between `providers` and `plugins`. Which do you prefer?
- Creation of jenkins plugin.


deploystream/apps/feature/lib.py
```diff
@@ -1,12 +1,37 @@
...
+    PLANNING_PLUGINS, SOURCE_CODE_PLUGINS, CI_PLUGINS
+)
+
+from models import Feature, Branch, PlanningInfo, CIInfo
+
+
 def get_feature_info(feature_id):
+    feature = Feature(feature_id)
     # First get any feature info from any management providers
+    for plugin in PLANNING_PLUGINS:
```
alexcouper, line 12: As mentioned in the PR comment - this `for` loop will need adapting to handle multiple plugins. At the moment it would override...

txels:
In (finally)

txels:
Big job man, congrats.

deploystream/apps/feature/lib.py
txels, line 12: I'm not sure it is a likely scenario to have multiple planning tools for the same project and feature... 
There may be a case where a project is migrating from one tool to another (which I have seen happen twice...) where some features are in the new tool and some still in the old one...
So it may be more a case of stopping at first success? And (as a user) sorting out your plugins by priority?

alexcouper:
Apologies that it's so long though! Next week I'll look at the remaining bullet points...

deploystream/apps/feature/lib.py
```diff
@@ -1,12 +1,37 @@
...
+        feature.planning_info = PlanningInfo(
+                                **plugin.get_feature_info(feature_id))
 
     # Then get any branch info from any source control providers
+    for plugin in SOURCE_CODE_PLUGINS:
+        for branch_data in plugin.get_repo_branches_involved(feature_id):
+            feature.branches.append(Branch(*branch_data))
```
txels: This breaks encapsulation, I'd rather have `feature.add_branch` than `feature.branches.append`.

deploystream/apps/feature/lib.py
```diff
@@ -1,12 +1,37 @@
...
 
     # Use that branch info, along with configuration regexes to create a
     # hierarchy of the branches involved in the feature.
+    feature.create_hierarchy_trees(app.config['HIERARCHY_REGEXES'])
```
txels, line 24: I still have in mind an idea about having some sort of `conventions` module that takes care of several project conventions (job naming, branching strategy, etc...). Either that, or each plugin takes care of conventions that apply to it. But not sure it belongs here.
Although this is good enough for now I guess.

deploystream/apps/feature/lib.py
txels: Note for later: check that we are not preventing having a project with different components each in a repo that uses different source control?

deploystream/apps/feature/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def __init__(self, repo_name, branch_name, latest_commit):
+        self.parent = None
+        self.children = []
+        self.siblings = []  # Will be needed in the cases where we have no
+                            # parent
+        self.ci_info = None
+        self.repo_name = repo_name
+        self.branch_name = branch_name
+        self.latest_commit = latest_commit
```
txels, line 96: we may need to store some reference to the plugin that provided this branch info

deploystream/apps/feature/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class PlanningInfo(object):
+
+    def __init__(self, title, id, url, feature_type, owner):
+        self.title = title
+        self.id = id
+        self.url = url
+        self.feature_type = feature_type
+        self.owner = owner
```
txels: same here, which plugin did instances of this come from? (in case of multi-planned projects)

deploystream/apps/feature/models.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class CIInfo(object):
```
txels: Better name here? `BuildInfo`?

deploystream/apps/feature/models.py
```diff
@@ -0,0 +1,43 @@
+class Feature(object):
```
txels, line 1: Now is the time for the docstrings...

deploystream/providers/__init__.py
```diff
@@ -0,0 +1,66 @@
+from importlib import import_module
+
+from zope.interface import classImplements, verify
```
txels: :)

In a separate note, in this case I prefer `from zope import interface` and `interface.classImplements`.

deploystream/providers/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+                    is_source_control_plugin,
+                    SOURCE_CODE_PLUGINS)
+
+    init_plugin_set(app.config['PLANNING_PLUGINS'],
+                    is_planning_plugin,
+                    PLANNING_PLUGINS)
+
+    init_plugin_set(app.config['CI_PLUGINS'],
+                    is_ci_plugin,
+                    CI_PLUGINS)
```
txels: I think it would be nicer to have this:

``` python
PLUGIN_TYPES = (
    ('SOURCE_CODE_PLUGINS', ISourceCodeControlPlugin),
    ('CI_PLUGINS', ICIPlugin),
    ...
)
```

...combined with a comment I will add below, and initialise the plugins looping through this structure.

deploystream/providers/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def _check_implements(cls, interface):
+    classImplements(cls, interface)
+    try:
+        verify.verifyClass(interface, cls)
+        return True
+    except Exception, e:
+        print e
+        return False
```
txels: Why not add a variation of this function in a superclass of the interfaces, so that you can do things like:

``` python
ISourceCodeControlPlugin.implemented_by(x)
```

deploystream/providers/__init__.py
```diff
@@ -0,0 +1,66 @@
+from importlib import import_module
+
+from zope.interface import classImplements, verify
+
+from interfaces import ISourceCodeControlPlugin, ICIPlugin, IPlanningPlugin
+
+PLANNING_PLUGINS = []
+SOURCE_CODE_PLUGINS = []
+CI_PLUGINS = []
```
txels: This could be a dict where the key is interface:

``` python
PLUGINS = {
    ISourceCodeControlPlugin: [],
    ICIPlugin: [],
    IPlanningPlugin: [],
}
```

deploystream/providers/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def is_planning_plugin(plugin):
+    "Check the plugin class given implements IPlanningPlugin."
+    return _check_implements(plugin, IPlanningPlugin)
+
+
+def is_ci_plugin(plugin):
+    "Check the plugin class given implements ICIPlugin."
+    return _check_implements(plugin, ICIPlugin)
```
txels: This code becomes redundant if you consider the rest of my recommendations.

deploystream/providers/git_plugin/plugin.py
```diff
@@ -0,0 +1,58 @@
...
+            The name of the repository to search for branches.
+
+        :param feature_id:
+            The id of the feature to look for in branches.
+
+        :param kwargs:
+            Expected to contain keys for:
+
+                - CODE_DIR: The directory in which repo ``repo_name`` can be
+                  found.
```
txels: Not sure why this `CODE_DIR` is a `kwarg` and not data passed in upon plugin instantiation.
Plugin configuration data should be passed via `__init__`.

deploystream/providers/interfaces/ci.py
```diff
@@ -0,0 +1,30 @@
...
+
+        :param branch:
+            The name of the branch to get build info for.
+
+        :param commit:
+            The commit of specific interest. This will be the commit that is at
+            the head of the given branch and repo_name. If there is no build
+            information for this commit it is acceptable to return build
+            information for the latest build - but it is important to include
+            the new commit in the return data.
```
txels: Ha, interesting. 
I am not too sure about the use case for passing the commit ID and not generating an error if it does not match the build.
What do you have in mind?

deploystream/providers/interfaces/ci.py
```diff
@@ -0,0 +1,30 @@
+from zope.interface import Interface
+
+
+class ICIPlugin(Interface):
+
+    def get_build_information(repo, branch, commit, **kwargs):
```
txels: One strange thing about zope interfaces is the missing `self` in the interface.

deploystream/providers/interfaces/ci.py
```diff
@@ -0,0 +1,30 @@
...
+
+        :param commit:
+            The commit of specific interest. This will be the commit that is at
+            the head of the given branch and repo_name. If there is no build
+            information for this commit it is acceptable to return build
+            information for the latest build - but it is important to include
+            the new commit in the return data.
+
+        :returns:
+            A dictionary containing keys and values for:
```
txels: at least...

deploystream/providers/interfaces/ci.py
```diff
@@ -0,0 +1,30 @@
...
+                - timestamp (of the build - a datetime object)
+                - result (one of "success", "failure", "unstable", "partial")
+                - commit (of the build)
+                - url
```
txels: Additional keys and values can be provided if the build system generates metrics such as violations, coverage...

deploystream/providers/interfaces/ci.py
```diff
@@ -0,0 +1,30 @@
...
+            The commit of specific interest. This will be the commit that is at
+            the head of the given branch and repo_name. If there is no build
+            information for this commit it is acceptable to return build
+            information for the latest build - but it is important to include
+            the new commit in the return data.
+
+        :returns:
+            A dictionary containing keys and values for:
+                - timestamp (of the build - a datetime object)
+                - result (one of "success", "failure", "unstable", "partial")
```
txels: I think we should have a core class `Result` with a string and a normalised status value (predefined constants). Similar to what we had in storyboard. We can use this for global build result, and also for any other metrics (see below). And possibly other plugins where a success/failure/partial success can make sense.

deploystream/providers/interfaces/planning.py
```diff
@@ -0,0 +1,27 @@
...
+            configuration of it. See :ref:`<configure_provider>` for more
+            information about configuring plugins.
+
+        :returns:
+            If found, a dictionary containing keys for at least:
+                - title
+                - id
+                - url
+                - feature_type
+                - owner
```
txels, line 17: - description (details) is probably another safe bet

deploystream/providers/interfaces/planning.py
```diff
@@ -0,0 +1,27 @@
...
+            by you then this will only be information you have added in the
+            configuration of it. See :ref:`<configure_provider>` for more
+            information about configuring plugins.
+
+        :returns:
+            If found, a dictionary containing keys for at least:
+                - title
+                - id
+                - url
+                - feature_type
```
txels, line 16: should we have normalised values?

deploystream/providers/interfaces/planning.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class IPlanningPlugin(Interface):
+
+    def get_feature_info(feature_id, **kwargs):
+        """
+        :param feature_id:
+            The identifier of a feature to be looked up.
+
+        :param **kwargs:
```
txels: Already commented elsewhere, I do not think an interface is the proper place to pass plugin configuration, which is implementation-dependent. It should be passed upon plugin instantiation.

deploystream/providers/interfaces/source_code_control.py
```diff
@@ -0,0 +1,42 @@
...
+class ISourceCodeControlPlugin(Interface):
+
+    def get_repo_branches_involved(feature_id, **kwargs):
+        """
+        Get the set of repo, branches involved in the given feature.
+
+        :param feature_id:
+            The id of the feature to look for in this all repos and branches.
+
+        :param **kwargs:
```
txels, line 13: same old

txels:
Thanks again for this massive piece of work. I, as usual, will be the one lagging behind...

txels:
BTW do you have in mind using something like this for API?: http://flask-restful.readthedocs.org/
I mention this because maybe you do not need to bother with views and templates for now, just provide an API.
I will be working on the assumption that the front-end bit relies just on a JSON API, maybe start writing bits based on sample dummy JSON data.

deploystream/apps/feature/lib.py
alexcouper: I've changed that to add_branch.

This should allow different source control versions working together - it is the plugins responsibility to bring back any branch from the plugin's associated version control system for the given feature id.

deploystream/apps/feature/lib.py
alexcouper, line 24: I'm not sure how that would pan out. My hope is that plugins are too low level to know or care about conventions and that the app asks the plugins simple questions based on conventions it knows about.

Perhaps we could discuss what the `conventions` module would look like at some point...

deploystream/apps/feature/models.py
alexcouper, line 96: Added

deploystream/apps/feature/models.py
alexcouper: Added

deploystream/providers/__init__.py
alexcouper: Changed

deploystream/apps/feature/models.py
alexcouper: Changed

deploystream/providers/interfaces/planning.py
txels, line 17: Oh and I think we need some normalised `status` such as defined, planned/current (from current sprint), in progress, completed, accepted, rejected... Maybe we can make a table of how this could work with the values we know from the different tools we've used (Rally, Pivotal, Sprint.ly, Jira...), it will be quite comprehensive with all these.
Each planning plugin would need to map the tool's values to normalised values so that we can do things like filter by status etc. and they work with data coming from different sources.

deploystream/apps/feature/models.py
alexcouper, line 1: Added a bunch in there...

deploystream/providers/interfaces/planning.py
```diff
@@ -0,0 +1,27 @@
...
+            If found, a dictionary containing keys for at least:
+                - title
+                - id
+                - url
+                - feature_type
+                - owner
+
+            ``None`` otherwise
+        """
+        pass
```
txels, line 22: I am thinking of implementing a feature list/detail as my first frontend example, getting this data from the planning tool (I will do a sprint.ly implementation). If I implement the feature list it will be a filtered list probably by status.

deploystream/providers/__init__.py
alexcouper: Not sure this is going to be possible. At least at first attempt it complains at me anyway - it needs to be a class method to make sense for us and zope interface says:

"InvalidInterface: Concrete attribute, is_implemented_by"

alexcouper: For now I'll move this function outside of **init**.py so at least this module is cleaner...

deploystream/providers/__init__.py
alexcouper: I've taken this approach now, but tweaked it slightly...

deploystream/providers/__init__.py
alexcouper: Taken this approach now, but tweaked it slightly to retain references to these lists...

deploystream/providers/__init__.py
alexcouper: removed

deploystream/providers/git_plugin/plugin.py
alexcouper: This is hangover from the days when plugins were expected to be modules. I'll change this.

alexcouper: Hmm although maybe not thinking about it....

It's conceivable that a single GitPlugin should be able to look up git repos in various locations - and only get that info at 'call' time.

Potentially we should make code_dir an explicit requirement in the interface though, rather than in kwargs. Thoughts?

deploystream/providers/interfaces/ci.py
alexcouper: I can see your point.

I wrote this assuming that if the latest build doesn't match the latest commit in the branch it is still valuable to know the build status.

So I wrote it assuming that the upstream caller would be responsible for figuring out if this build matched the job.

## Why pass the commit here in the first place then?

In the standard jenkins job there'd be no need to pass in "commit" to this function - it could just return the latest build status and the commit for which it built.

**BUT**

In my new found pet-love - namely the github PR builder in jenkins - we would get build information a different way. We would need a github CI plugin (which I have already written) which goes off to the given commit and gets the commit status from it. Inside that status is the success/failure etc.

What I could do is change this interface to be only taking `repo`, `branch` and returning `status`, `commit`. Then in the github builder we would need to somehow specify that it wants to know about the commit it should be looking for in its configuration...

deploystream/providers/interfaces/ci.py
alexcouper: I've added this.

At the moment any additional key will actually fail though - as we aren't accepting **kwargs in the `class BuildInfo` (formerly `class CIInfo`).

Do you think we should change that?

deploystream/providers/interfaces/ci.py
alexcouper: I've added a to do for that. In my mind it would be up to our end to convert this result into a `Result` to avoid having plugins importing a `Result` class from us...

deploystream/providers/interfaces/planning.py
alexcouper, line 17: Any chance of adding that as an issue on the repo? ;)

deploystream/providers/interfaces/planning.py
alexcouper, line 16: Probably, yes.

deploystream/providers/git_plugin/plugin.py
alexcouper: Actually I think the plugin should manage all that itself - so will add a constructor for this...

deploystream/providers/interfaces/planning.py
alexcouper: Removed all these **kwargs sections from the plugins...

alexcouper:
I've made all the changes I _think_ need to be made now - but do alert me if I've missed something you think is important.

I've added a TODO.txt where we can for now keep track of outstanding things (I think we used this approach on the original pretenders repo and it worked quite well).

deploystream/providers/interfaces/planning.py
txels, line 17: Great suggestion, there you go: #9

deploystream/providers/interfaces/planning.py
txels, line 16: There you go: #10

TODO.txt
```diff
@@ -0,0 +1,19 @@
...
+ - Finish Git plugin - to make it look across a group of repos
+ - Code that handles the creation of hierarchical branch trees
+ - git plugin code to use hierarchical trees to determine merge statuses
+
+
+Unclaimed
+=========
+ - Code to handle multiple plugins - we'll need to merge data between them I suspect.
+ - Code that passes per-plugin settings to the plugins.
+ - There's inconsistency in the naming between providers and plugins. Which do you prefer?
```
txels, line 13: I prefer `providers` for code that implements a given service, and `plugins` for add-on (possibly 3rd party) components that can potentially be installed dynamically (e.g. the way you can implement Jenkins plugins).

A plugin may implement one or more providers (e.g. a Github/Bitbucket plugin may implement some source control features plus some planning features - via its support for issues, plus some team information provider, if we ever support that...).

txels:
Beautiful.

PR #7, opened 2013-03-03
Initial AngularJS structure

txels:
This contains some initial structure of the Angular-based frontend.
- It has been seeded with the 'angular-seed' template project, so there is not much genuine work in here ;)
- It contains the basic scaffolding for unit and end to end tests as well.

For now it just renders some hardcoded data.

Next steps:
- Adapt it to provide a view that talks to the backend via an actual API call
- Provide a real API endpoint in the backend so that we have a first complete end-to-end running system.


frontend/app/css/app.css
```diff
@@ -0,0 +1,30 @@
...
+
+.menu > li:before {
+  content: "|";
+  padding-right: 0.3em;
+}
+
+.menu > li:nth-child(1):before {
+  content: "";
+  padding: 0;
+}
```
txels, line 30: Most of this will be removed, it was part of the angular-seed project

frontend/app/js/controllers.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+/* Controllers */
+
+
+function FeatureListCtrl($scope) {
+    $scope.features = [
+        {id: '812', title: 'Build something good'},
+        {id: '350', title: 'Check it out'}
```
txels, line 9: hardcoded data...

alexcouper:
I'm afraid I won't be able to look at this until tomorrow evening at the earliest! 

Quick glance looks good though!

alexcouper:
44000+ lines. Wow.

Merging.

txels:
Hey, most of these 44000 lines are JS libraries...
I will get rid of the ones we don't need, and use minified versions for the others.
I've been taking a look at http://twitter.github.com/bower/ for dependency management in the frontend world (aka pip for js/css). It may work for us.

PR #12, opened 2013-03-03
Use defaultdict to store various data elements in feature

txels:
The main idea behind this is to make certain base behaviour in `Feature` extensible without modification in the future, enabling different types of plugins/interfaces/providers.

This is only a showcase, if you like the idea I would suggest to also change the plugin registry to use a similar pattern.

### DIFF
--- a/deploystream/apps/feature/models.py
+++ b/deploystream/apps/feature/models.py
@@ -1,3 +1,6 @@
+from collections import defaultdict
+
+
 class Feature(object):
     """
     The class used for encapsulating ``Feature`` data across repos & branches.
@@ -21,7 +24,7 @@ class Feature(object):
     """
     def __init__(self, id):
         self.id = id
-        self.branches = []
+        self.info = defaultdict(list)
         self.planning_info = None
         self.trees = []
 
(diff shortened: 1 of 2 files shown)


deploystream/apps/feature/models.py
```diff
@@ -21,7 +24,7 @@ class Feature(object):
     """
     def __init__(self, id):
         self.id = id
-        self.branches = []
+        self.info = defaultdict(list)
         self.planning_info = None
         self.trees = []
```
txels, line 29: trees are a candidate to have a `Tree` class and be part of `info`.

alexcouper:
It's elegant and I'm not against it - but I'm not certain whether it's something we want to do.

I'm a fan of things being explicit when they are known. We know for example that a feature has branches. It's not going to have many other lists containing instances of one type. And to me generalizing often has the side effect of losing information about intention.

eg. In your example, a new user looking at the `Feature` class would have no idea what kinds of things are expected to be stored in info. Can it store anything? well, yes. But the system at large would pretty much only ever be storing branches in there.

Perhaps there's more mileage in the plugin version though where we _might_ have n types of plugins. Although TBH even there I imagine we'll only have 3 in which case generalising isn't going to save us much...

What do you think?

txels:
I was thinking extensibility, but the lack of explorability of the interface is a fair point. 

I also found it a bit boring to have to create an `add` method for every piece of info we want to add to a story. I am thinking that planning and build info can also belong in here (I am not sure if build info belongs only in branches), and in the future deploy info, even marketing feedback info (A/B testing data) or any other crazy thing some developer using this may like to include etc.

In terms of extensibility the nice thing is that you could add arbitrary new data types, interfaces and providers... when you think of future extension via plugins this is an important point.

For explorability, we can add an explicit property `branches` that returns `self.info[Branch]` etc. That could even be generated at runtime... it would look again a bit like magic - but it could be found via the Python shell ;)

I will give this some additional thought.

alexcouper:
I do like it - I just feel that maybe it's too early in the day to be doing this kind of optimization

txels:
Just for the fun, I added creation of properties at runtime, so that you can see `f.branches` in the Python shell... only after you have already added one branch :P

alexcouper:
Mind if I close this PR for now - or were you hoping for a merge on this?

txels:
Will close myself, but it will be on the back of my head for later...

PR #13, opened 2013-03-03
Backend is an API that serves JSON data

txels:
The backend becomes an API that will serve JSON data. This is a starting point.
Most templating will be in the frontend side of things handled by Angular.

So we probably need to find the best pattern to serve JSON data from Flask.
I am sure there are better ways, or if this one is good for us we could just make this pattern it into a decorator `@as_json`.


alexcouper:
This'll do for now - although I'm completely unsure as to the performance of jsonpickle.

PR #14, opened 2013-03-03
Add git branch finder and test

alexcouper:
@txels 

This is now a full-blown PR


deploystream/settings.py
```diff
@@ -1,6 +1,17 @@
...
 BUILD_INFO_PLUGINS = []
 
 HIERARCHY_REGEXES = []
+
+
+PLUGIN_CONFIG = {
+    'GitPlugin': {
+        'code_dir': None
+    }
+}
```
txels: What about having the config in the plugin list? As in plugin list contains tuples with class name (str) and config (dict).
BTW what made you keep all the separate lists of plugins, instead of a single data structure?

txels:
Re: your question - what about creating a repo in github for the sole purpose of using it in `deploystream` tests? E.g. deploystream-test-repo. As long as we only clone it and never push to it we are fine I guess, any people could run any number of tests in parallel.

But if we want to push to it, I guess the github API supports creation of repos as well. You can create one with a random name.

deploystream/settings.py
alexcouper: The motivation behind the separate lists was to do with attempting to make customizing configuration easier. But actually, it only reduces the pain without solving it. 

But since we're thinking about db-driven per-user configuration this is probably a non-issue now.

alexcouper:
I've created a dummy repo - I'll use that in the tests. For now there's no pushing to the repo anyway.

deploystream/settings.py
```diff
@@ -1,8 +1,8 @@
-from local_settings import GITHUB_CONFIG
+from local_settings import GITHUB_CONFIG, GIT_CONFIG
 
 
 SOURCE_CODE_PLUGINS = [
-    ('deploystream.providers.git_plugin.plugin.GitPlugin', {}),
+    ('deploystream.providers.git_provider.GitProvider', GIT_CONFIG),
```
alexcouper, line 5: The reason for the git_provider name rather than git is because git is also the name of a library that we use.

alexcouper:
@txels: this is ready for review now...

deploystream/settings.py
txels, line 5: yeah I remember that problem from gdfabric and storyboard

deploystream/providers/git_provider/__init__.py
```diff
@@ -20,7 +23,17 @@ def get_repo_branches_involved(self, feature_id):
                 1: branch name
                 2: latest commit
         """
-        return []  # no info available for now
+        # Every folder in side self.code_dir that is a git repo will be looked
```
txels: `inside`?

deploystream/providers/git_provider/__init__.py
```diff
@@ -20,7 +23,17 @@ def get_repo_branches_involved(self, feature_id):
...
+        # at
+        repo_branches = []
+        for repo_name in os.listdir(self.code_dir):
+            repo_location = join(self.code_dir, repo_name)
+            if exists(join(repo_location, ".git")):
+                branches = self.get_branches_involved(repo_location,
+                                                      feature_id)
+                repo_branches.extend([(repo_name, ) + branch
+                                            for branch in branches])
```
txels: weird indent, is this valid PEP8?

deploystream/providers/git_provider/__init__.py
```diff
@@ -42,8 +55,9 @@ def get_branches_involved(self, repo_location, feature_id):
                         .format(repo_location=repo_location))
         remote = git.remote.Remote(repo, 'origin')
         affected = []
+        lower_feature = feature_id.lower()
```
txels: sounds like lowly feature :P

tests/test_providers/data/__init__.py
```diff
@@ -0,0 +1 @@
+# This folder is used to house data for tests.
```
txels: I guess this is never going to be a package, why `__init__.py`? can use use `.empty` or similar?
And add to `.gitignore`?

tests/test_providers/test_git_provider.py
```diff
@@ -0,0 +1,39 @@
...
+from nose.tools import assert_equal
+
+from deploystream.providers.git_provider import GitProvider
+
+DUMMY_CODE_DIR = join(dirname(__file__), 'data')
+
+
+def ensure_dummy_clone_available():
+    """
+    Check that we have access to pretenderers' dummyrepo
```
txels: them ones who pretend even more...

tests/test_providers/test_git_provider.py
```diff
@@ -0,0 +1,39 @@
...
+def test_git_plugin_finds_branches_across_repos():
+    """
+    Test that the GitPlugin finds branches in repos in the dir given.
+
+    Clone the dummyrepo into the data folder if not already there.
+
+    The data in this test is found by looking at the dummyrepo and getting
+    the branch names and latest commit of any branches that match "FeAtUrE".
+    """
+    ensure_dummy_clone_available()
```
txels: you could use nose's `with_setup`: https://nose.readthedocs.org/en/latest/testing_tools.html#nose.tools.with_setup

deploystream/providers/git_provider/__init__.py
```diff
@@ -42,8 +55,9 @@ def get_branches_involved(self, repo_location, feature_id):
...
         for remote_ref in remote.refs:
-            if feature_id in remote_ref.remote_head:
+            if lower_feature in remote_ref.remote_head.lower():
```
txels: I guess this is a purposefully naive implementation of matching for now...
Are you planning to define a way to match branches according to defined rules/conventions? Such as `<user>/<feature_id>-<text>` or `<feature_id><.*>` or include things like a `develop` or `staging` branch in the mix etc...?

Maybe it is premature optimization, but I think if we keep VCS-service-specific code separate from process-related code (e.g. how the team approaches branching) it will make the architecture better, and adaptors to other VCS easier to write. In this case, I would rather keep the adaptors to VCS dumb and just tell them to give me a list of branches that match one in a list of regexes or sthg like that.

txels: BTW this service could be implemented with the Github provider as well... and the code is actually simpler. Which makes a stronger case for keeping the source control provider API so that implementations can be as dumb as possible.

I guess as long as you do not need local copies of your code (which AFAIR we only used for deploying, and deployment is very specific to each company) we could do everything with the github API.

txels:
out

deploystream/providers/git_provider/__init__.py
alexcouper: Re: everything via github api.

I still think its worth having the git provider. Reason being that it allows the system to work against any git provider rather than just github.

I think it's worth adding a github source code provider though. Added as #19.

Re: Regexes.
The user story in my mind for get_branches_involved is #22. I'm planning on changing the code here now to allow the plugin to take a `feature_breakup` regex and a `branch_finder` regex. 

Hopefully that matches what you're after too...

deploystream/providers/git_provider/__init__.py
alexcouper: Apparently so - at least I don't get an error on mine. Reformatted anyway.

tests/test_providers/data/__init__.py
alexcouper: Removed

deploystream/providers/git_provider/__init__.py
alexcouper: fixed

tests/test_providers/test_git_provider.py
alexcouper: fixed

tests/test_providers/test_git_provider.py
alexcouper: fixed

alexcouper:
I've added the regexing.

If this doesn't cover a use case that you are thinking of then perhaps we should have a chat about it?

txels:
OK maybe we should chat or I should add some example, because what you did is very different from what I had in mind.

alexcouper:
I'm up for chatting about it tomorrow (sat) morning. Does that work for you?

alexcouper:
OK, well seems like you're not around online at the moment and I need to head off shortly.

At the moment the core code calls `get_repo_branches_involved` on a VCS provider and at that point only knows the feature id. The purpose of this is so that we can get back all the information to do with a feature. I'm unsure then why we need to worry about staging, or develop or users or anything else.

So a question is "who should own the breaking up of a feature id into parts that can be matched in branch names". And I guess that could live up in the core code, and then the `get_repo_branches_involved` could instead take as arguments the keywords used in the breaking up code.
eg. 

```
# Config
FEATURE_BREAKUP = "(?P<project>[a-zA-Z]+)-?(?P<id>[0-9]+)" 

# Core code
match = re.search(config.FEATURE_BREAKUP, feature_id)
vcs_plugin.get_repo_branches_involved(**match.groupdict())

# VCS provider
def get_repo_branches_involved(self, **kwargs):
    # kwargs contains project and id keys in this case
    regex_matcher = self.template.format(**kwargs)
    # ...
```

and then the provider itself is configured with the template regex for branch finding. eg `".*{project}{id}.*"`. In this way retrieving would be configurable and decoupled from the process logic.

But I think fundamentally we need to get on the same page over which part of the system is responsible for what - and I'm very unclear as to the purpose of your desire for `staging`, `develop` or `name/` matching.

txels:
That is the gist of it. The VCS provider code should not know about features, only branches.

In the scheme I had in mind, an upper layer would call it only with a list of strings to match for a specific case. I would make it look like:

``` python
# VCS 
def get_branches(self, regexes=None):
    """
    regexes == None - return all branches
    regexes == list of strings: branches whose name matches any of the regexes
    """
    ...
```

`regexes` could be a list, or a single regex using `|` (I think the list would be more useful later, as we could use it to express branch hierarchy when looking for things like branches that are [un]merged on the way up [e.g. th way up for GD would be user_dev -> story -> rc -> staging -> master, somewhere else this could be story_branch -> develop -> master, etc]).

On a specific case (story 347) it could get called with:

``` python
regexes = ["^staging$", "^rc_us347", "^us347", "^us347.*_dev_.*"]  # GD
regexes = ["^develop$", "^\w+/347-.*"]  # RBX
```

The concrete list of matchers would be built by a higher level, not the VCS plugin itself, where project rules or conventions (such as branch naming and structure, etc) would be defined.

I do not mean this is how it _should_ be, just how I was thinking it would end up being.

txels:
BTW I'm happy to merge this as it is, I think we will figure out the right APIs when implementing end to end stories (as happened with feature list). Let's define an end-to-end story that uses this code.

PR #16, opened 2013-03-04
Add a planning provider for GitHub

txels:
@alexcouper this is just so that you know what I am doing and where I am heading, and add your comments before I stray too far.
### What I have thought and done
- I have thought that a good use case for deploystream is to help us manage the development of itself. So the first implementation of a planning service is going to be Github issues. Then I can start getting a list of features to render in the frontend etc.
- While working on it I realised that we hadn't added the possibility to retrieve a list of features from anywhere, so I put it in the Planning service. 
- There is a new view to list all features under `/features`. You will need to configure local settings (ah, the pain of local configuration).

``` python
GITHUB_CONFIG = {
    'repositories': [('pretenders', 'deploystream')],
    'username': 'alexcouper',
    'password': '...',
}
```
### Missing
- Tests, documentation...
- Properly encode unicode in UTF-8
### What I think we need to do additionally
- When doing the above I realised that the features themselves will need to store a reference to the planning plugin where they came from, and realised that if we are always going to retrieve features from some planning tool, what we had labelled as planning info maybe are just regular fields on `Feature`.
- Properly define how we want configuration to work. That was pretty weak in storyboard as well. We will need to separate user-specific from global settings etc.

### DIFF
--- a/deploystream/apps/feature/lib.py
+++ b/deploystream/apps/feature/lib.py
@@ -1,9 +1,22 @@
+import itertools
+
 from deploystream import app
 from deploystream.providers import (
     PLANNING_PLUGINS, SOURCE_CODE_PLUGINS, BUILD_INFO_PLUGINS
 )
 
-from models import Feature, Branch, PlanningInfo, BuildInfo
+from .models import Feature, Branch, PlanningInfo, BuildInfo
+
+
+def get_all_features():
+    features = []
+
+    for plugin in PLANNING_PLUGINS:
+        features += plugin.get_features()
+
+    # features = itertools.chain([plugin.get_features() for plugin in PLANNING_PLUGINS])
+
+    return features
 
 
 def get_feature_info(feature_id):
--- a/deploystream/apps/feature/lib.py
+++ b/deploystream/apps/feature/lib.py
@@ -7,7 +7,9 @@ from models import Feature, Branch, PlanningInfo, BuildInfo
 
 
 def get_feature_info(feature_id):
-    feature = Feature(feature_id)
+    # TODO: we need to provide a way to identify where the feature came from,
+    # possibly by namespacing the feature_id parameter. For now we use None
+    feature = Feature(None, feature_id)
     # First get any feature info from any management providers
     for plugin in PLANNING_PLUGINS:
         feature.planning_info = PlanningInfo(
--- a/deploystream/apps/feature/models.py
+++ b/deploystream/apps/feature/models.py
@@ -22,7 +22,7 @@ class Feature(object):
                             branches are merged into their parents etc.
     """
     def __init__(self, plugin, id):
-        self.plugin = plugin
+        self._plugin = plugin
         self.id = id
         self.branches = []
         self.planning_info = None
@@ -60,7 +60,7 @@ class PlanningInfo(object):
         self.feature_type = feature_type
         self.owner = owner
         self.description = description
-        self.plugin = plugin
+        self._plugin = plugin
         self.extras = kwargs
 
 
@@ -98,7 +98,7 @@ class Branch(object):
         self.repo_name = repo_name
         self.branch_name = branch_name
         self.latest_commit = latest_commit
-        self.plugin = plugin
+        self._plugin = plugin
 
 
 class BuildInfo(object):
@@ -123,4 +123,4 @@ class BuildInfo(object):
         self.result = result
         self.commit = commit
         self.url = url
-        self.plugin = plugin
+        self._plugin = plugin
--- a/deploystream/apps/feature/models.py
+++ b/deploystream/apps/feature/models.py
@@ -61,7 +61,7 @@ class PlanningInfo(object):
         self.owner = owner
         self.description = description
         self._plugin = plugin
-        self.extras = kwargs
+        self._extras = kwargs
 
 
 class Branch(object):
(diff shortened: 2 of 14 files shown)


deploystream/apps/feature/models.py
```diff
@@ -49,8 +52,8 @@ class PlanningInfo(object):
         ``plugin``          - The plugin that provided all of these values.
     """
 
-    def __init__(self, title, id, url, feature_type, owner, description,
-                 plugin):
+    def __init__(self, plugin, id, title, feature_type='story',
+                 owner=None, description=None, url=None, **kwargs):
```
txels, line 56: Added some defaults, plus arbitrary fields that different plugins may implement (and that make it simpler to just unpack JSON data)

deploystream/providers/interfaces/planning.py
```diff
@@ -3,6 +3,18 @@
 
 class IPlanningPlugin(interface.Interface):
 
+    def get_features(**filters):
```
txels, line 6: We will need to normalise the filtering parameters.

deploystream/providers/__init__.py
```diff
@@ -21,9 +17,8 @@
 
 def get_plugin_class(path):
     "Given a path to a class import the module and return the class"
-    index_pos = path.rindex('.')
-    mod_path, class_name = path[:index_pos], path[index_pos + 1:]
-    mod = import_module(mod_path)
+    mod_path, class_name = path.rsplit('.', 1)
+    mod = __import__(mod_path, globals(), locals(), [class_name])
```
txels, line 21: this is able to load classes from `__init__.py`, which the previous code couldn't

alexcouper:
### Configuration

In terms of plugin-dependent config settings I think we need to look at it entirely from the point of view of someone using deploystream.

One thing I had been thinking would be for them to be able to provide an ini style config file or directory of files that defines all the plugins being used and their configurations. Not sure what you think about that though...

### Get all features.

You're right we did need a call like this.

We may need to provide some way of reducing the set brought back as Rally/Jira/X's APIs are probably all pretty laggy. Either paginating calls or getting only what is relevant.

In the predecessor to deploystream a function existed `get_all_features_in_current_iteration` which obviously wouldn't apply to github issues but which would mean we're not bringing back lots of useless stuff.

### Feature **is** PlanningInfo

Yep, this makes sense

### Other

I like the idea of using deploystream to manage itself early on. 
I assume that this is not merge-ready based on your comment and that you'll be adding to this PR. If not, let me know!

Nice work!

txels:
I changed things to use token authentication in GIthub, and added a script to get the token for you for now.

txels:
You can merge this whenever you please, so that it does not get too large.
We can record pending stuff as issues and take it from there.

deploystream/apps/feature/lib.py
```diff
@@ -1,13 +1,28 @@
...
+from .models import Feature, Branch, PlanningInfo, BuildInfo
+
+
+def get_all_features():
+    features = []
+
+    for plugin in PLANNING_PLUGINS:
+        features += plugin.get_features()
+
+    # features = itertools.chain([plugin.get_features() for plugin in PLANNING_PLUGINS])
```
alexcouper, line 17: Remove line?

deploystream/apps/feature/models.py
```diff
@@ -19,7 +21,8 @@ class Feature(object):
                             the front end to display information about what
                             branches are merged into their parents etc.
     """
-    def __init__(self, id):
+    def __init__(self, plugin, id):
+        self._plugin = plugin
```
alexcouper, line 25: Why the "secrecy" ?

deploystream/lib/transforms.py
```diff
@@ -0,0 +1,52 @@
...
+    if isinstance(data, unicode):
+        return data.encode('utf-8')
+    if isinstance(data, basestring) or isinstance(data, int):
+        return data
+    elif isinstance(data, list) or isinstance(data, tuple):
+        return [nativify(x) for x in data]
+    elif isinstance(data, dict):
+        return {
+            k: nativify(v)
+            for k, v in data.items() if not k.startswith('_')
```
alexcouper, line 17: Won't this will mean we lose the `extras` from PlanningInfo?

I think one of these if statements should check if data has a nativify method too and return the calling of that?

deploystream/lib/transforms.py
```diff
@@ -0,0 +1,52 @@
...
+        }
+    elif data is None:
+        return 'null'
+    elif hasattr(data, '__dict__'):
+        return nativify(data.__dict__)
+    else:
+        return '"{0}"'.format(data)
+
+
+def remap(original, keymap):
```
alexcouper, line 27: nice

alexcouper:
[out]

As you say, we should raise the rest as issues

deploystream/apps/feature/models.py
txels, line 25: For now the JSON serialisation we have does not render fields starting with `_`. It is a simple way to filter JSON output until we come up with something better, if needed.

deploystream/lib/transforms.py
txels, line 17: I did this on purpose to hide that data for now from the JSON API.
We can write a story to define how we properly want JSON to work.

alexcouper, line 17: OK fair enough

alexcouper:
Merged

deploystream/lib/transforms.py
txels, line 17: I call this our poor man's JSON.

PR #18, opened 2013-03-06
Frontend rename app and use resource for Feature

txels:
This starts using an angular `resource`, which is a proxy object to a RESTful API server resource. I also properly renamed the app.

At some point I need to tie this up with the Flask backend, so that the whole JS stuff is served as static data. Do you have an idea of how you want to set up Flask to serve static files during development?

### DIFF
--- a/frontend/app/js/services.js
+++ b/frontend/app/js/services.js
@@ -6,5 +6,17 @@
 
 // Demonstrate how to register services
 // In this case it is a simple value service.
-angular.module('myApp.services', [])
-.value('version', '0.1');
+angular.module('deploystream.services', ['ngResource'])
+    .value('version', '0.1')
+    .factory('Feature', function ($resource) {
+        return $resource(
+            'features/:featureId',
+            {},
+            {
+                query: {
+                    method: 'GET',
+                    params: {featureId: 'all.json'},
+                    isArray: true
+                }
+            });
+    });
--- a/frontend/app/partials/features.html
+++ b/frontend/app/partials/features.html
@@ -1,6 +1,6 @@
 <p>These are your project's features:</p>
 <ul>
     <li ng-repeat="feature in features">
-        {{feature.id}}: {{feature.title}}
+        {{feature.id}}: {{feature.planning_info.title}}
     </li>
 </ul>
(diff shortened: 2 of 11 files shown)


frontend/app/js/services.js
```diff
@@ -6,5 +6,17 @@
 
 // Demonstrate how to register services
 // In this case it is a simple value service.
-angular.module('myApp.services', [])
-.value('version', '0.1');
+angular.module('deploystream.services', ['ngResource'])
+    .value('version', '0.1')
+    .factory('Feature', function ($resource) {
```
txels, line 11: This here would be the resource proper.

frontend/app/partials/features.html
```diff
@@ -1,6 +1,6 @@
 <p>These are your project's features:</p>
 <ul>
     <li ng-repeat="feature in features">
-        {{feature.id}}: {{feature.title}}
+        {{feature.id}}: {{feature.planning_info.title}}
```
txels, line 4: this will change once we merge `Feature` and `PlanningInfo`

alexcouper:
Flask handles static files for you during development. Just pop them in a static/ folder 

http://flask.pocoo.org/docs/quickstart/#static-files

frontend/app/js/services.js
```diff
@@ -6,5 +6,17 @@
...
+        return $resource(
+            'features/:featureId',
+            {},
+            {
+                query: {
+                    method: 'GET',
+                    params: {featureId: 'all.json'},
```
alexcouper, line 18: why "all.json". Shouldn't this be :featureId somehow (my angular < )

alexcouper:
I assume this needs to be tied up to be served by flask before it works properly...

I'll merge - but I'm interested to know about that all.json thing...

frontend/app/js/services.js
txels, line 18: This is the value that it will use when there is no feature ID, i.e. to retrieve the resource list. For now I faked it to be `/features/all.json` so that it was served off a local file I generated from the actual response of the backend.
On my next step I will tie both pieces together.

Issue #23, opened 2013-03-10
Set up project to serve static files, and serve the angular frontend this way


txels:
In order to be able to use a simple configuration during development, serve static files via the Flask development server. This will require some shifting of frontend files around...

PR #24, opened 2013-03-10
Link together the angular and flask sides

txels:
First step of solving #23.

Frontend tests and scripts will need to be adjusted to match the new structure, but for now a full end-to-end app (that lists open features) can be browsed at http://localhost:5000/

Another pending thing is to ~~serve the homepage on `/` from a flask view, and~~ turn the static `index*.html` into proper jinja templates in `templates`.


txels:
Now we finally have some sort of end to end stack to build on. This could be end of _iteration 0_ and maybe we can structure backlog stories around end to end features from now on.

local_settings.py
```diff
@@ -2,7 +2,11 @@
 from github_auth import TOKEN
 
 GITHUB_CONFIG = {
-    'repositories': [('pretenders', 'deploystream')],
+    'repositories': [
+        ('pretenders', 'deploystream'),
+        ('pretenders', 'pretenders'),
+        ('txels', 'autojenkins'),
+    ],
```
txels, line 9: this file shouldn't be under VCS (at least under this name) but it serves to illustrate that this works with multiple sources of features

alexcouper:
Looks good.

I'm wondering whether we shouldn't be stricter on the interface between providers and core.
In the code above we add a default of `None` for `project` in a `Feature`.

I'm wondering whether we should explicitly state that is required - especially since we use it on the template. If a provider doesn't want to give it, then perhaps it should be up to it to provide a default value.

deploystream/providers/github/__init__.py
```diff
@@ -37,8 +37,9 @@ def get_features(self, **filters):
 
         for owner, repo in self.repositories:
             ghrepo = self.github.repository(owner, repo)
+            project = '{0}/{1}'.format(owner, repo)
             for issue in ghrepo.iter_issues(**filters):
-                feature = Feature(self, issue.number)
+                feature = Feature(self, issue.number, project=project)
```
alexcouper: I'm not sure I like the fact that this provider imports and uses`Feature` and `PlanningInfo` classes - as it means the plugin knows more than it should and breaks the idea of providers having simple "give args, return dicts" methods. To me, this method should return a list of dictionaries that are converted by the core processing logic. What do you think?

local_settings.py
alexcouper, line 9: Yeah - hopefully we'll do user-specific configuration soon...

alexcouper:
I'm also unsure about the models in `features/models.py` accepting `**kwargs` as I can't see when we would use them in a good way....but I won't bring this issue into this PR... :)

deploystream/providers/github/__init__.py
txels: I'm not sure about returning dicts. 

The point in favour of the `dict` approach is to reduce the _framework_ feel of the project, where one can build a plugin with no imports to the project. (I believe this only applies to third party developers of plugins that are not contributing to this repo, but building a 3rd party package that one can install as an extension to it - we're quite far from that, if we ever get there...).

A point against it is the fact that provider developers will have to rely on extra documentation (the keys that must be included in the returned dict), which means we will have to keep two pieces of code in two different modules in sync. That doesn't sound very DRY to me.

I'm probably missing something, so what do you think? Do you believe the first argument has more in its favour than the second?

txels:
Re: mandatory fields, I agree. We should enforce a number of fields. I made it optional just because I was adding it and it wasn't there before, but it should be mandatory, as a number of others.

Re extra `**kwargs`. TBH the main reason why I added that was to save me from writing filtering code in the github plugin, but then I also thought: maybe we want plugins to be not just backend but also frontend/templating code. This would allow a developer of a provider to be able to display additional fields in the frontend that are specific to that provider and relevant to the user. So in case of doubt I took the lazy approach, which is writing the least code required to implement my need (or extreme programmers' "simplest thing that can possibly work").

txels:
BTW I will now be getting rid of `PlanningInfo` and put everything in `Feature`.

deploystream/providers/github/__init__.py
alexcouper: I don't think you're missing anything - those two points are the main pros & cons - along with compatibility issues for third party plugins (eg PlanningInfo loses an argument and then all the 3rd party providers break).

On the surface it does seem to only matter if we ever get to the point of having 3rd party devs - which as you say is a long way away - but i'm wondering whether it exposes something wrong in our API, or the way we're using it. Perhaps the provider should be thinking more low level and the bulk of get_features done at a higher level or something. But even if more is moved away from the Provider we'll still hit this issue of "to import or not to import". In short I don't actually have an answer.

I'll merge this now and I'll try to find time tomorrow to have a look at how other people do this kind of thing.

alexcouper:
Oh - just seen your BTW - I'll leave you to it. Let me know when you want a merge...

deploystream/providers/github/__init__.py
txels: I removed the imports, as once flattening PlanningInfo into Feature it didn't make much of a difference in the code, and for now it stays consistent with the assumptions in `get_feature_info` tests

txels:
I just realised one thing about the `**kwargs`: it allows for forwards-compatibility. Meaning a plugin written for a newer version of `deploystream` that supports additional arguments e.g. for `Feature` will still work on an older version.

txels:
You may merge this at your leisure. I will do some work on tests and documentation but this needn't delay the merge, and you can work on a sync'ed version. Would be nice to have a chat on _what next_, a pity we didn't seem to have very compatible schedules this weekend.

alexcouper:
Merging

Issue #27, opened 2013-03-12
Login using github credentials via github oauth dance.


alexcouper:

txels:
BTW not sure whether you realised the script `pr-for-issue` allows you to turn an issue into a PR (which you cannot do via the web interface of github). Thus reducing the number of "duplicated" issued and the need to remember to close issues.

alexcouper:
Ooh nice. Didn't see that, will use in future...

PR #36, opened 2013-03-17
Story/27/alex

alexcouper:
@txels - I decided to stick with Flask-Oauth in the end, and I'm using the certifi certificates to make sure we get the latest.

If it stops working it _should_ be a case of updating requirements (providing certifi is kept up to date).


alexcouper:
This is a fix for #27

Issue #37, opened 2013-03-17
Change plugin initialisation


alexcouper:
We currently init all the plugins in the constructor of deploystream with all config info.

This has to change as we are wanting multiple users.

either.

a) Continue with init there, but pass all configuration into each method call.
b) Create the plugins as they're needed, for each request, discarding them afterwards.
c) Something else?

alexcouper:
Probably related to #17

alexcouper:
I'm going to have to take care of this as part of #28 

I'm going with (b).

txels:
Why not session? Initialise user's plugins upon login...

Issue #8, opened 2013-03-03
Providers or plugins? Providers


alexcouper:
@txels:

Which do we want to call them?

We currently have a package called providers and we call them plugins pretty much everywhere else...

txels:
I realise I had only replied to this in an already closed PR. For the record:

I prefer `providers` for code that implements a given service, and `plugins` for add-on (possibly 3rd party) components that can potentially be installed dynamically (e.g. the way you can implement Jenkins plugins).

A plugin may implement one or more providers (e.g. a Github/Bitbucket plugin may implement some source control features plus some planning features - via its support for issues, plus some team information provider, if we ever support that...).

Since I think it is too early to get into implementing some sort of dynamic plugin system, we may want to use `provider` exclusively for now, if you agree with my definitions.

PR #38, opened 2013-03-17
Story/28/alex

alexcouper:
This is for #28, #37 and #8  and should be merged after #36 ( #27 )
## Settings changes:
- While we wait for accounts, I've set up a `USER_SPECIFIC_INFO` variable in settings.py to simulate user specific plugin configuration.
- `PROVIDERS` is now a list of provider paths with no care for their types.
## Provider changes:
- We used to initialise all providers, check for types and store them in lists at kick off. Now we're just going to import the relevant classes at the start and upon each request we'll use the relevant ones for that request. 
- Currently a "relevant" provider is defined as one whose name is present as a key in the `'provider_config'` key of `USER_SPECIFIC_INFO`.
- Relevant providers are instantiated with the config settings at request time and passed in to the appropriate function (`get_feature_info`/`get_all_features`) who is responsible for checking whether they're implementing the right thing before calling.
- If a provider defines `oauth_required`, the relevant key will be looked up in the current session tokens and returned. The github token resides at "github".
## Future thoughts:
- I'm thinking that the oauth stuff could actually just be another type of provider. So adding integration to a new oauth-required service would be pretty easy.
- Also wondering whether `get_all_features`/`get_feature_info` should just call the functions blindly without checking interface details, catching attribute errors.

If you agree with either of the _future thoughts_, I'll add them as issues to be done in this iteration...


alexcouper:
I've pulled in `develop` post your merge of the other PR in order to get a decent diff for you.

Slightly smaller now...

txels:
Yes, both _future thoughts_ sound reasonable.
As per initialisation of providers, maybe upon session creation (login), and settings changes (whenever we enable that) would be a good place.

deploystream/providers/interfaces/build_info.py
```diff
@@ -1,7 +1,13 @@
...
 
 
-class IBuildInfoPlugin(interface.Interface):
+class IBuildInfoProvider(interface.Interface):
+
+    name = interface.Attribute(
+                "The name the provider will be referred to in configs etc.")
+    oauth_token_required = interface.Attribute(
+                "If an oauth token is required, the name of it as defined by "
+                "the oauth provider.")
```
txels, line 10: these could be part of a base `ProviderInterface` to avoid repetition.

deploystream/apps/feature/lib.py
```diff
@@ -1,61 +1,77 @@
...
 
 
-def get_all_features():
+def get_all_features(providers):
     """
     Collect all features from the configured providers.
     """
     all_features = []
 
-    for provider in PLANNING_PLUGINS:
```
txels, line 15: I had the idea that even if ending up with a single structure, we would classify the providers by interface upon load, and still end up with different provider lists by type.

I.e. when loading a plugin, you figure out which interfaces it provides, and add it to a dictionary such as:

``` python
PROVIDERS = {
    IPlanningProvider: [ <github>, <jira> ],
    ISourceCodeControlProvider: [ <git> ],
    IBuildInfoProvider: [ <jenkins> ],
    IOAuthProvider: [ <github> ],
}
```

Although it will be clearer what to do once we let users configure their providers from the list of available (installed) providers.

This line would then be:

`for provider in providers[IPlanningProvider]`

deploystream/apps/feature/lib.py
```diff
@@ -1,61 +1,77 @@
...
     # This needs rewriting according to previous paragraph. For now:
     # Only one management provider should know about this feature,
     # so we stop on first success
-    for plugin in PLANNING_PLUGINS:
-        feature = Feature(plugin, None, **plugin.get_feature_info(feature_id))
-        if feature:
-            break
+    feature = None
+    for provider in providers:
+        if is_planning_provider(provider):
```
txels, line 41: similarly, `for provider in providers[IPlanningProvider]:`...

deploystream/apps/oauth/github.py
```diff
@@ -18,13 +18,22 @@
...
+
+
 @github_oauth.tokengetter
 def get_github_token(token=None):
     """Function used in the flask-oauth library to get a token.
 
     For now just get it from the session, or return ``None``.
     """
-    return session.get('github_token')
+    return (session.get('tokens', {}).get('github'), '')
```
txels, line 36: can you make this 2 lines of code, for clarity?

deploystream/apps/oauth/github.py
```diff
@@ -44,7 +53,8 @@ def github_authorized(resp):
     if resp is None:
         flash(u'You denied the request to sign in.')
         return redirect(next_url)
-    session['github_token'] = (resp['access_token'], '')
+
+    add_session_token(session, resp['access_token'], "github")
```
txels, line 57: DRY strings that denote session keys? And put somewhere central enough to have all session keys in a single place, for easier conflict management.

`SESS_OAUTH_TOKEN='github'`

deploystream/providers/decorators.py
```diff
@@ -0,0 +1,20 @@
+from functools import wraps
+
+from flask import session
+
+from deploystream import app
+from deploystream.providers import get_providers
+
+
+def needs_providers(func):
```
txels, line 9: Nice

deploystream/providers/git_provider/__init__.py
```diff
@@ -7,6 +7,8 @@
 
 class GitProvider(object):
 
+    name = 'git'
```
txels, line 10: `oaut_token_required` should be part of the interface now, as declared below?

deploystream/providers/interfaces/build_info.py
txels, line 10: on second thought, do these really need to be part of the interface? E.g. the way I implemented it, name is not mandatory, and defaults to class name. Can this be the same with oauth, and default to `None`?

txels:
Looks good, only some minor comments.

deploystream/apps/oauth/github.py
txels, line 57: on second thoughts, this is provider-specific session data, so I guess for these keys we should use sthg like `<provider-name>.<key>`, such as `"github.oauth-token"`.

txels:
I decided to merge this since none of the comments was actually blocking, and this is much more useful than `develop` already - I realised when I noticed I am creating my new branches off this one ;)

deploystream/providers/interfaces/build_info.py
alexcouper, line 10: These `interface.Attribute` types aren't inforced but are just a placeholder for documentation.

Or at least `verify.verifyClass(interface, cls)` returns `True` despite their absence.

deploystream/apps/feature/lib.py
alexcouper, line 15: OK cool - this can be done.

alexcouper, line 15: I think this would pretty much negate the second "future thoughts" above. I'll assume that we want to continue with interfaces and do this change rather than the "just ask the plugin for stuff" approach.

PR #35, opened 2013-03-18
Display logged in username at top of the page.

txels:
Besides the logged in user's name, this includes a dropdown menu with options for logout (functioning) and settings (a placeholder that we can always comment-out).

I was playing with bootstrap customisation and fonts to give deploystream a more distinctive look, and I cheekily included these in the PR. Let me know how you like the new fonts - we can always revert to the standard helveticas if you don't like these...


alexcouper:
Why does the diff not show any of the changes to the actual template files?

txels:
Hmm that is github being annoying when there are too many diff lines.
I will rewind the PR to before adding the fonts, and do a separate PR for that.

txels:
Cleared up. Will send another PR with fonts and colors, mainly for you to judge on aesthetics rather than code.

alexcouper:
Merging.

PR #45, opened 2013-03-20
Carles: playing with fonts and colors

txels:
Github is still not very smart at filtering the diff, it even ignores `.gitattributes`.

But this is basically for you to test visually.
You can run a diff of this branch with origin/develop locally and it will properly exclude svg files from the diff.


alexcouper:
The front page looks a bit funny with the tan colour when compared with the blue button.

But I think this is us gradually moving away from an obviously bootstrapped theme so I'm cool with merging.

PR #44, opened 2013-03-21
Make changes following code review

alexcouper:
https://github.com/pretenders/deploystream/pull/38

### DIFF
--- a/deploystream/apps/feature/lib.py
+++ b/deploystream/apps/feature/lib.py
@@ -1,6 +1,6 @@
 from deploystream import app
 from deploystream.providers.interfaces import (
-    is_planning_provider, is_build_info_provider, is_source_code_provider)
+    IBuildInfoProvider, IPlanningProvider, ISourceCodeControlProvider)
 from .models import Branch, BuildInfo, Feature
 
 
@@ -10,10 +10,9 @@ def get_all_features(providers):
     """
     all_features = []
 
-    for provider in providers:
-        if is_planning_provider(provider):
-            for feature in provider.get_features():
-                all_features.append(Feature(provider, **feature))
+    for provider in providers[IPlanningProvider]:
+        for feature in provider.get_features():
+            all_features.append(Feature(provider, **feature))
 
     return all_features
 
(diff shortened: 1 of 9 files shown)


deploystream/apps/feature/lib.py
```diff
@@ -37,41 +36,37 @@ def get_feature_info(feature_id, providers):
...
-        if is_planning_provider(provider):
-            feature = Feature(provider, None,
-                              **provider.get_feature_info(feature_id))
-            if feature:
-                break
+    for provider in providers[IPlanningProvider]:
+        feature = Feature(provider, None,
+                          **provider.get_feature_info(feature_id))
+        if feature:
+            break
```
txels, line 43: There is an interesting point re: how we deal with PRs. If a PR in Github is just an issue coming from the planning interface, then in my case I may have information for a single issue coming from two sources that needs to be aggregated in some way. 

If we treat a PR just as additional "decoration" for a feature (as we can do with branches and jobs) then it can be dealt with at the next stage, when we add info to a feature.

Maybe we will end up needing a generic `IFeatureDecorator` interface that just does arbitrary stuff to a feature (as it gets passed through a pipeline of these providers) to support more flexible tooling.

deploystream/apps/oauth/github.py
```diff
@@ -2,6 +2,7 @@
 from flask_oauth import OAuth
 
 from deploystream import app
+from deploystream.apps.oauth import get_oauth_token, set_oauth_token
```
txels, line 5: so much oauth... wouldn't this read better:

``` python
from deploystream.apps import oauth

oauth.get_token()
oauth.set_token()
```

?

deploystream/apps/oauth/__init__.py
```diff
@@ -13,3 +13,13 @@ def ensure_certifi_certs_installed():
     if not filecmp.cmp(source, destination, shallow=False):
```
txels, line 13: Out of curiosity, why is this in `apps` and other stuff not in `apps`? What is your rule of thumb?
It does not have views, which would have been my pseudo-rule...

deploystream/providers/__init__.py
```diff
@@ -18,23 +25,33 @@ def get_providers(config_dict, session):
...
             print ("WARNING: A provider wanted a token but we didn't have one")
             pass
-        providers.append(provider_class(**kwargs))
+
+        provider = provider_class(**kwargs)
+
+        for iface in [IBuildInfoProvider, IPlanningProvider,
+                      ISourceCodeControlProvider]:
+            if is_implementation(provider.__class__, iface):
+                providers[iface].append(provider)
```
txels, line 54: would be nice to issue some sort of warning for providers that didn't match any interface? log message?

txels:
Looks like you started using the `pr-for-issue` script ;)

txels:
Looks good to merge. As mentioned by email, would be nice to have a chat re: how we deal with PRs.

deploystream/apps/oauth/__init__.py
alexcouper, line 13: Oauth stuff has views, and this function is only needed because of the oauth stuff.

So I guess I follow that pseudo-rule but probably got this one wrong :)

Given that it is called by **init** of deploystream it probably should live elsewhere TBH.

deploystream/apps/oauth/github.py
alexcouper, line 5: Changed in the other PR.

PR #29, opened 2013-03-21
View a list of issues from all my repos in github (via github provider)

alexcouper:


alexcouper:
This should be looked at after https://github.com/pretenders/deploystream/pull/44 as I branched from that to make it.

I tried doing what the issue describes, but found that it was really slow. (in pseudo code)

```
get all the repos that the user has
get all the organisations that the user is part of
for each
   get all the repos

loop through all repos, find all issues and return them
```

it took >60 seconds to respond (i'm a member of gittip org which has a lot of issues).

I also can't really think of a use case where you'd want to look across organizations at the same time, so I left this level of configuration in the system.

txels:
Yes, make these kind of things configuration. I am doing something similar with sprintly options as they arise... at some point we may want to define a more general approach to plugin settings, for now stick things in a dict as we've done so far.

txels:
You probably want to rebase this off `develop` for Github to properly show the changes.

alexcouper:
_rebased_

alexcouper:
Assuming Travis likes this, it's ready to be merged.

You'll want to make changes to your configurations though...

deploystream/providers/github/__init__.py
```diff
@@ -22,41 +22,46 @@ class GithubProvider(object):
...
-    def __init__(self, token, repositories, **kwargs):
+    def __init__(self, token, organization=None, **kwargs):
         """
         Initialise the provider by giving it GitHub credentials and repos.
 
-        :param repositories:
-            A list of tuples containing (<owner>, <name>) that identify
-            a repository in GitHub
+        :param organization:
+            The name of the organization who's repository issues should be
```
txels, line 30: whose

deploystream/providers/github/__init__.py
```diff
@@ -22,41 +22,46 @@ class GithubProvider(object):
...
+            identified in GitHub. If ``None`` then the authenticated
+            user's issues will be tracked.
         """
         self.github = github3.login(token=token)
-        self.repositories = repositories
+        if not organization:
```
txels, line 35: I know this implementation more or less suits what we discussed for iteration 0 (personal-config-free, almost). But I realised this doesn't really match my usage profile, which is either my company context (where I want to see `rockabox` repos, and this could work), or my personal context (where I have own repos + pretenders repos). A list of organizations here would work better for me. 

But in fact in the real use case when we start hosting this we will want to offer the user a list of personal repos and organizations to choose from, in a similar way as to how Travis does it. So I won't worry about this too much for now until we start doing proper settings.

txels:
Works for now. We will address personal configuration properly later.

PR #43, opened 2013-03-22
Add sprint.ly planning provider

### LINKED ISSUE #47: Guarantee the execution order of providers
When I have several providers of the same type, I cannot define the order in which they should run.

E.g. my `sprintly` provider always runs before `github`, regardless of settings.

txels:
http://help.sprint.ly/knowledgebase/topics/15784-api

Possibly using: https://github.com/txels/apitopy

This PR additionally fixes #47


deploystream/providers/__init__.py
```diff
@@ -16,12 +16,12 @@ def get_provider_class(path):
...
 
-def get_providers(config_dict, session):
+def get_providers(configs, session):
     """
     Get appropriate providers for the given session.
 
-    :param config_dict:
-        A dictionary of provider name to configuration.
+    :param configs:
+        A list of tuples with provider name and configuration.
```
alexcouper, line 24: That's better

deploystream/providers/sprintly/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+            **feature.assigned_to)
+    feature = transforms.remap(feature, FEATURE_MAP)
+    return feature
+
+
+class SprintlyProvider(object):
+    """
+    An implementation of the planning provider that gets stories from Sprint.ly
+    """
+    interface.implements(IPlanningProvider)
```
alexcouper, line 32: We check for this anyway, so not sure the benefit of having it. I guess for explicitness?

deploystream/providers/sprintly/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __init__(self, user, token, current, **kwargs):
+        """
+        Initialise by providing credentials and project IDs.
+
+        Load a list of available projects.
+        """
+        self.api = Api('https://sprint.ly/api/',
+                       (user, token), verify_ssl_cert=False, suffix='.json')
+        self.projects = self.api.products()
```
alexcouper, line 43: Purely FYI to make sure we're on the same page - this is the same as having on line 52 `for project in self.api.products()` in the sense that we don't "save" any time by putting it here.

Currently all providers are instantiated on every request, so as a result all the projects are retrieved each time - wherever this code is put.

This is something we may want to change - to allow a provider to store some information in the session, or in some cache. But that's for another day - just thought I should raise it in case you were under the impression that providers were created once per session and only get_features() was called each request.

txels, line 43: oh that is a bug, I was not intending to do the request twice.

txels, line 43: BTW this needs lots of testing.
I am wondering what our approach to this should be, as it is such a clear use case for pretenders...

alexcouper, line 43: I like your thinking

deploystream/providers/sprintly/__init__.py
txels, line 32: Yes, I think this way the code makes a clear statement. This will be more apparent once we have providers that implement multiple services.

I believe the point of us checking it the way we do was because we did not want plugin writers to have to import anything from the project, but I don't see why we as project maintainers should do the same.

deploystream/providers/sprintly/__init__.py
txels, line 43: Oh sorry now I understand your comment.
I think it is good to keep the initialisation in `__init__` anyway, as I also think we will soon move to the environment you describe, with more things cached or in the session.

PR #49, opened 2013-03-22
Some restructuring of scripts and settings

txels:
The motivation behind this is threefold:
- I was not very happy with the project structure and mixed location of scripts and settings. I've tried to put all scripts and settings in the `scripts` and `config` directories.
- I do not like the fact of having files that do not live under version control and I am moving away from that pattern as much as possible in my current django projects. In particular this applies to `local_settings`. I am now only keeping personal secrets out of the way (and ideally we should get these from environment vars instead of python files).
- I am now running deploystream (TBR) for personal stuff and already sharing it with people in the office, and I found it a pain to manage the two different sets of repos (when we discuss settings in proper detail, we may want to introduce the concept of "contexts")

Now you can add your own settings in `config` too, e.g. `config/local_settings_alex.py` etc. In order to run:

```
CONFIG=alex ./scripts/runserver.sh
```

For tests, use `CONFIG=test nosetests` or `./scripts/test`


txels:
There is one single test failing in Travis due to the fact that we try to clone a repo from Github and within the Travis server that fails with `Permission denied (publickey).`.

txels:
\o/ Travis!

config/local_settings_carles.py
```diff
@@ -0,0 +1,15 @@
+GITHUB_CONFIG = {
+    'repositories': [
+        ('pretenders', 'deploystream'),
+        ('pretenders', 'pretenders'),
```
alexcouper, line 4: This will need updating after #29

config/local_settings_carles_rbx.py
```diff
@@ -0,0 +1,21 @@
+from sprintly_auth import TOKEN
+
+GITHUB_CONFIG = {
+    'repositories': [
+        ('rockabox', 'rbx_web'),
+        ('rockabox', 'conf'),
+        ('rockabox', 'youtube-frames'),
```
alexcouper, line 7: Cool that you've already started using this at work!

config/settings.py
```diff
@@ -0,0 +1,47 @@
...
+    LOCAL_SETTINGS = 'local_settings_{0}'.format(CONFIG)
+    mod = __import__(APP_PACKAGE, globals(), locals(), [LOCAL_SETTINGS], -1)
+    submod = getattr(mod, LOCAL_SETTINGS)
+    where = globals()
+    for attr in dir(submod):
+        if not attr.startswith('__'):
+            where[attr] = getattr(submod, attr)
+            print("imported " + attr)
+except (AttributeError, ImportError):
+    import sys
```
alexcouper, line 19: Might be nice to also print original traceback in case the ImportError or AttributeError occurs in the local settings file.

deploystream/__init__.py
```diff
@@ -5,18 +5,20 @@
...
 # Set configuration defaults from deploystream settings.
-app.config.from_pyfile(join(APP_DIR, 'settings.py'))
+app.config.from_pyfile(join(CONFIG_DIR, 'settings.py'))
 
 # Override with anything in DEPLOYSTREAM_SETTINGS
 if environ.get("DEPLOYSTREAM_SETTINGS"):
     app.config.from_envvar("DEPLOYSTREAM_SETTINGS")
 
 try:
+    # TODO find a more appropriate way to initialise these (env vars?)
```
alexcouper, line 21: Add a github issue for iteration 1?

deploystream/providers/github/__init__.py
```diff
@@ -54,6 +58,9 @@ def get_features(self, **filters):
                     issue_info['owner'] = owner.login
                 features.append(issue_info)
 
+        # sort by putting PRs first, stories second
+        features = sorted(features, key=lambda f: f['type'] == 'story')
```
alexcouper, line 62: Not sure this should really be in the provider. If we really want it here, think it should be parameter driven through the filters argument. 

If this is just for display purposes, can't Angular do this for us?

alexcouper:
out for now

deploystream/providers/github/__init__.py
txels, line 62: you're absolutely right, this was a quick hack that I wanted to move to configuration. But I think this type of default/preferred sorting still belongs to backend. We want our API to provide a sorted list of relevant stories based on user configuration.

alexcouper:
Gonna merge this and add the additional issues to the iteration. Partly because they're not show stoppers and partly so that my pr is green in travis :)

config/local_settings_carles_rbx.py
txels, line 7: Yes, even some colleagues eventually use it ;)

deploystream/__init__.py
txels, line 21: saw you just did.

PR #56, opened 2013-04-01
Carles/loading icon

txels:
I finally figured out how to add a loading icon in Angular.
Much awaited feature...

### DIFF
--- a/config/local_settings_carles_rbx.py
+++ b/config/local_settings_carles_rbx.py
@@ -1,6 +1,7 @@
 from sprintly_auth import TOKEN
 
 GITHUB_CONFIG = {
+    'organization': 'rockabox',
     'repositories': [
         ('rockabox', 'rbx_web'),
         ('rockabox', 'conf'),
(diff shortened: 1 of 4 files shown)


config/local_settings_carles_rbx.py
```diff
@@ -1,6 +1,7 @@
 from sprintly_auth import TOKEN
 
 GITHUB_CONFIG = {
+    'organization': 'rockabox',
     'repositories': [
```
alexcouper, line 5: You can get rid of this now too - don't think it does anything anymore...

txels, line 5: I keep it just in case it comes back ;)
I find that I am typically interested in a only a subset of repos at a time... so we need to refine the approach.

If you remember, the idea we choose to implement initially of "all of the user's repos" was aimed to avoid having configuration on iteration 0. But this no longer holds, as you found out it was way too slow... So we will have to discuss whether we do some sort of user settings for a first public release.

PR #58, opened 2013-04-06
Choose and implement a color scheme

txels:
An initial state. You can play a bit with it by commenting in and out the colour palettes in colours.less (and shuffling colour numbers) and recompiling (`./scripts/compile-css.sh`).

Additionally I have switched from `recess` to `lessc` to compile less files (`recess --compile` is just a wrapper around `lessc` and I've found it limits command-line options and obscures error messages).


deploystream/static/less/colours.less
```diff
@@ -0,0 +1,34 @@
+// I like your smile Palette ---------------
+@color1: #B3CC57;
+@color2: #ECF081;
+@color3: #FFBE40;
+@color4: #EF746F;
+@color5: #AB3E5B;
```
txels, line 6: I am cheekily using this one as it seems to give me more colourful results. 
But I haven't played with the citrus colours enough yet.
Anyway I am trying to make it trivial to replace the initial set of 5 colours, and let all the others derive from those.

alexcouper:
Something has messed up the layout of the front page:

Screen Shot 2013-04-07 at 11 38 44 AM

alexcouper:
It's a shame that npm doesn't work like pip in the sense that it always downloads and installs regardless of whether it is needed (ie run the get_dependencies script twice and it'll take almost as long the second time).

alexcouper:
I'm happy to merge this for now, and we can take a look at the layout issue later on.

txels:
Open #62 to fix this.
As per your comment on NPM, avoiding re-installs is one of the things I would like the `webdeps` tool to do. Bower has the same problem.

PR #67, opened 2013-04-09
Make JS e2e tests pass again

txels:
I've changed the test slightly so that it passes now.

I'd like to get these tests back into shape by making them do something useful, but that will mean using `pretenders`. It will be simplest if it is hosted somewhere, that way we may be able to run these tests within Travis.


alexcouper:
I see...think we should do something about that soon...

PR #48, opened 2013-04-10
Make oauth a provider type

alexcouper:
This PR includes:
- Slight rearrange of files
- Addition of a OAuth provider interface
- Additional tests to ensure that Github is an OAuth provider
- Changes to enforcing interface variables so that attributes are now enforced.
- Auto creation on start up of all OAuth provider urls for use as needed.
- Addition of some simple logic that does the following:
  - If a provider requires OAuth and that OAuth isn't present, redirect to go and perform the OAuth.
- A test to prove above

The redirect works in theory (and you can test it by logging out and going to `/features`. Angular isn't currently expecting any kind of redirect response when it does a $resource of `/features` so in reality this doesn't really help much.

The reason for that addition is so that when I finish the JIRA implementation, you'll be redirected off to confirm the new app only when you need to (rather than having to manually click somewhere to start it off).

Once this has been merged (and I find some more time ;) ), I'll be continuing with #65. I guess I'll have to deal with the Angular redirect in that.


deploystream/__init__.py
```diff
@@ -21,8 +21,8 @@
     # TODO find a more appropriate way to initialise these (env vars?)
     from github_auth import APP_ID, APP_SECRET
     app.config.update({
-        "GITHUB_APP_ID": APP_ID,
-        "GITHUB_APP_SECRET": APP_SECRET,
+        "github_APP_ID": APP_ID,
+        "github_APP_SECRET": APP_SECRET,
```
txels: this case mix looks really weird

deploystream/apps/oauth/github.py
```diff
@@ -1,59 +0,0 @@
-from flask import session, redirect, flash, request, url_for
```
txels, line 1: this file is still referenced from within `deploystream/__init__.py`

deploystream/apps/oauth/views.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def configure_route(oauth_name, oauth_data):
+    """
+    Configure some app routes for the given oauth provider.
+
+    It is assumed that config exists for::
+
+        <oauth_name>_APP_ID
+        <oauth_name>_APP_SECRET
```
txels: could we _uppercase_ oauth_name, or better yet use it as a key to a dict that contains ID and SECRET?

txels: Shouldn't this config somehow live in the provider's settings?

deploystream/apps/oauth/views.py
```diff
@@ -0,0 +1,90 @@
...
+        username = user.data['login']
+        session['username'] = username
+
+    return redirect(next_url)
+
+
+@app.route('/login')
+def login():
+    "Handler for calls to login via github."
+    return start_token_processing('github', islogin=True)
```
txels, line 76: shouldn't this (`'github'`) come from a setting?

alexcouper, line 76: At the moment we only want logins to be via github - or at least if we want to generalise that then we need to support that visually on the front page too..

deploystream/lib/__init__.py
```diff
@@ -0,0 +1,15 @@
+from os.path import join, dirname
+import shutil
+import filecmp
+
+import certifi
+import httplib2
+
+
+def ensure_certifi_certs_installed():
```
txels, line 9: We should document here why we do this. Since there are no references to `httplib2` elsewhere in the code, it is not clear why we are patching it, i.e. which of our dependencies actually uses it. So that we can get rid of this code if sometime we get rid of that dependency...

deploystream/apps/oauth/github.py
alexcouper, line 1: Weird that travis passed...

deploystream/__init__.py
alexcouper: I agree

deploystream/lib/__init__.py
alexcouper, line 9: Yep we should ( and to prove your point, i'll need to remind myself which library it was! )

deploystream/providers/__init__.py
```diff
@@ -37,13 +39,15 @@ def get_providers(configs, session):
         if config:
             kwargs.update(config)
         try:
-            kwargs['token'] = oauth.get_token(
+            if provider_class.oauth_token_required:
+                token = oauth.get_token(
                                     session,
                                     provider_class.oauth_token_required)
```
txels: now that I read this code sometime after it was originally written, `oauth_token_required` seems to suggest a boolean value.

deploystream/providers/interfaces/__init__.py
```diff
@@ -1,27 +1,45 @@
 from zope import interface as zinterface
 from zope.interface import verify
-
 # These imports are here for convenience, to be re-imported by others
 from build_info import IBuildInfoProvider
 from planning import IPlanningProvider
 from source_code_control import ISourceCodeControlProvider
+from oauth import IOAuthProvider
```
txels: can we use dot for relative imports? Else this is a bit dangerous.

``` python
from .build_info import IBuildInfoProvider
...etc
```

deploystream/apps/oauth/views.py
alexcouper: It's one of the things we haven't really sorted in terms of settings.

The ID and secret will be deploystream wide, regardless of user - where as the provider settings are all currently user dependent.

I think this is why you suggested in `deploystream/__init__.py` to potentially have them as env vars.

We can certainly uppercase, but I'm not sure about the placement of these into a provider.

deploystream/providers/__init__.py
alexcouper: how about `oauth_token`?

deploystream/providers/interfaces/__init__.py
alexcouper: Yep, sure

tests/test_oauth/test_get_token.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def get_oauth_data(cls):
+        return {
+            'base_url': 'https://api.github.com/',
+            'request_token_url': None,
+            'access_token_url': 'http://access_token_url',
+            'authorize_url': 'http://auth_url',
+            'request_token_params': {
+                'scope': 'user,repo'
```
txels, line 21: This is not the place in the code, but we actually may not need `user` in scope for accessing the github API. It may be off-putting to some users, and we do not really need to write profile information.
http://developer.github.com/v3/oauth/#scopes

tests/test_providers/test_github_provider.py
```diff
@@ -42,3 +44,9 @@ def test_get_features(github3):
     assert_equal(features[1]['id'], '2')
     assert_equal(features[1]['type'], 'story')
     assert_equal(features[1]['owner'], '')
+
+
+@patch('deploystream.providers.github.github3')
+def test_implements_expected_interfaces(github3):
```
txels: `github3` is not used, maybe replace with `_`

txels:
The PR-for-issue makes things a bit confusing, as it mistakes the PR date for the issue creation date.
It also sets the Issue creator as the PR author.
Maybe we just have to modify that script a bit.

deploystream/providers/__init__.py
txels: That would do. Or `oauth_token_name`. Your call

deploystream/apps/oauth/views.py
txels: Yes you are right these are site-wide now, and it will only be Github login for now anyway.

alexcouper:
I'll make these and any other change suggestions over the weekend given 10 mins.

alexcouper:
I've made all the above changes. Let me know if there's anything else :)

deploystream/providers/sprintly/__init__.py
```diff
@@ -34,6 +34,7 @@ class SprintlyProvider(object):
     """
     interface.implements(IPlanningProvider)
     name = 'sprintly'
+    oauth_token_name = None
```
txels, line 37: No biggie, but I'd rather we only needed to add this field for providers that do actually require it.

deploystream/providers/github/__init__.py
```diff
@@ -73,3 +73,15 @@ def get_features(self, **filters):
...
+
+    @classmethod
+    def get_oauth_data(self):
+        return {
+            'base_url': 'https://api.github.com/',
+            'request_token_url': None,
+            'access_token_url': 'https://github.com/login/oauth/access_token',
+            'authorize_url': 'https://github.com/login/oauth/authorize',
+            'request_token_params': {
+                'scope': 'repo'
```
txels, line 85: yup this is where I meant :)

deploystream/providers/sprintly/__init__.py
alexcouper, line 37: Me too. Doesn't seem like Zope interfaces support optional attributes so it
is either not included or mandatory :(

txels, line 37: Ha, that is the whole point of interfaces, enforcement :P

Issue #65, opened 2013-04-07
Add JIRA planning provider


alexcouper:

alexcouper:
Part way through this, but it now has a slight dependency on #48 so going to solve that first...

PR #68, opened 2013-04-23
Story/65/alex

alexcouper:
This attacks #65 (finally)

### DIFF
--- a/config/local_settings_alex.py
+++ b/config/local_settings_alex.py
@@ -13,8 +13,8 @@ GIT_CONFIG = {
 
 JIRA_CONFIG = {
     'url': '',
-    # 'user': None,
-    # 'password': None,
+    'user': None,
+    'password': None,
 }
 
 
--- a/config/local_settings_alex.py
+++ b/config/local_settings_alex.py
@@ -9,3 +9,17 @@ GIT_CONFIG = {
     'feature_breakup_regex': "(?P<project>[a-zA-Z]+)-?(?P<id>[0-9]+)",
     'branch_finder_template': ".*(?i){project}.*"
 }
+
+
+JIRA_CONFIG = {
+    'url': 'https://glassesdirect.atlassian.net',
+    'user': None,
+    'password': None,
+}
+
+
+try:
+    from non_github_settings import JIRA_CONFIG
+except ImportError:
+    print "Failed to import from non_github_settings"
+    raise
--- a/config/local_settings_alex.py
+++ b/config/local_settings_alex.py
@@ -12,9 +12,9 @@ GIT_CONFIG = {
 
 
 JIRA_CONFIG = {
-    'url': 'https://glassesdirect.atlassian.net',
-    'user': None,
-    'password': None,
+    'url': '',
+    # 'user': None,
+    # 'password': None,
 }
 
 
--- a/deploystream/lib/transforms.py
+++ b/deploystream/lib/transforms.py
@@ -24,7 +24,7 @@ def nativify(data):
         return '"{0}"'.format(data)
 
 
-def remap(original, keymap):
+def remap(original, keymap, defaults=None):
     """
     Return a new dictionary using data from original mapped using keymap.
 
@@ -50,6 +50,8 @@ def remap(original, keymap):
         Only keys defined in keymap will be moved across to the new dictionary.
 
     """
+    if defaults is None:
+        defaults = {}
     new_dict = {}
 
     for find_key, new_key in keymap.iteritems():
@@ -57,6 +59,8 @@ def remap(original, keymap):
             # Get the nested value of the composite key
             found_value = original
             for sub_key in find_key:
+                if not found_value:
+                    break
                 found_value = found_value[sub_key]
         else:
             found_value = original.get(find_key)
--- a/deploystream/lib/transforms.py
+++ b/deploystream/lib/transforms.py
@@ -26,7 +26,7 @@ def nativify(data):
 
 def remap(original, keymap):
     """
-    Generate a new dictionary by mapping certain keys to others.
+    Return a new dictionary using data from original mapped using keymap.
 
     This is a base adaptation layer to simplify translating various
     field names as provided by 3rd party services so that they are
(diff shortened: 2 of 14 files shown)


config/local_settings_alex.py
```diff
@@ -9,3 +9,17 @@
...
+
+JIRA_CONFIG = {
+    'url': '',
+    'user': None,
+    'password': None,
+}
+
+
+try:
+    from non_github_settings import JIRA_CONFIG
```
txels, line 22: :)

deploystream/lib/transforms.py
```diff
@@ -35,18 +35,39 @@ def remap(original, keymap):
...
         API).
+
     :param keymap:
         A dictionary mapping original field name => transformation.
         Transformation can be a single string, representing the new
         name for the field, if only the name changes, or a tuple
         containing `(new_name, value_mapping)`.
+        Original field name can be a single string, representing the location
+        in the original dict to find the value, or a tuple of keys if the value
+        is nested.
```
txels, line 46: nice

txels, line 46: `:param defaults:` ?

deploystream/lib/transforms.py
```diff
@@ -35,18 +35,39 @@ def remap(original, keymap):
...
+
+    .. note::
+
+        Only keys defined in keymap will be moved across to the new dictionary.
```
txels: Oh. This is a change in semantics. Why this change?

tests/test_lib/test_mappings.py
```diff
@@ -34,3 +34,15 @@ def test_name_and_value_remap(self):
...
+
+    def test_nested_name_remap(self):
+        """
+        Test nested field names can be mapped.
+        """
+        original = {"somekey": {"identifier": 123, "other_data": 445}}
+        mapping = {('somekey', 'identifier'): "id"}
+        expected = {"id": 123}
+
+        mapped = remap(original, mapping)
```
txels: should include a test case with the additional argument `defaults`.

deploystream/providers/jira/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+                - title
+                - id
+                - url
+                - feature_type
+                - owner
+                - description
+
+            ``None`` otherwise
+        """
+        pass
```
txels: Maybe we should start using `NotImplementedError` instead of `pass` (me guilty)

tests/test_providers/test_jira_provider.py
```diff
@@ -0,0 +1,39 @@
...
+
+    features = jira_provider.get_features()
+
+    assert_equals(features[0]['title'], "Perform update")
+    assert_equals(features[0]['id'], "TD-193")
+    assert_equals(features[0]['url'],
+        "https://something.atlassian.net/rest/api/2/issue/46385")
+    assert_equals(features[0]['type'], "Sub-task")
+    assert_equals(features[0]['owner'], "alex.couper")
+    assert_equals(features[0]['project'], "Superior Project")
```
txels, line 39: of course :)

tests/test_providers/test_jira_provider.py
```diff
@@ -0,0 +1,39 @@
...
+from tests import load_fixture
+
+ISSUE_RESPONSE_JSON = load_fixture('jira_issues.json')
+
+
+def test_is_planning_provider():
+    assert_true(is_implementation(JiraProvider, IPlanningProvider))
+
+
+@patch('jira.client.requests')
```
txels, line 16: Ha, this is extreme whitebox, having to look at the source code of an external library in order to write a test case... :P

txels:
Out.

deploystream/lib/transforms.py
alexcouper: We could bring it back if you want. I changed it because JIRA brings back a huge amount of data with each issue so it seemed a "waste" to copy across keys and nested dicts that were never going to be used.

deploystream/providers/jira/__init__.py
alexcouper: Changed

tests/test_lib/test_mappings.py
alexcouper: I've removed this additional argument as it seems to have been hangover from an earlier refactor attempt

tests/test_providers/test_jira_provider.py
alexcouper, line 16: I know :( - but a prime pretenders refactor candidate.

And it's made me aware of this kind of use-case for pretenders more generally: When using any third party library to connect to an API. 

The only 2 non-pretender ways are to: a) do this extreme whitebox, or b) do similar extreme white box of creating the third party objects and returning them.

deploystream/lib/transforms.py
alexcouper, line 46: Removed the unused `defaults` param

deploystream/lib/transforms.py
txels: I was thinking that eventually some day the plugins/providers we would build would have a frontend as well as backend bits, and the frontend bits would be able to display additional info that may not be deploystream-normalised data, but provider-specific. There is nothing in this PR that prevents that, but it makes it a little tedious to have to change code in two places for that.

While I was testing Github and sprintly I found it useful to get everything that was coming through, and find additional useful fields that I had not initially considered.

That said, there is nothing wrong with being totally explicit here, but it means I will have to test that github and sprintly still work, which I doubt as I was relying on the fact that some basic fields have typically the same names we expect in deploystream (title, description, etc...).

deploystream/providers/jira/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+from deploystream.lib import transforms
+
+# Map fields from the Jira API to deploystream names
+FEATURE_MAP = {
+    ('fields', 'summary'): 'title',
+    'key': 'id',
+    'self': 'url',
+    ('fields', 'issuetype', 'name'): 'type',
+    ('fields', 'assignee', 'name'): ('owner', {None: ""}),
+    ('fields', 'project', 'name'): 'project',
```
txels, line 13: Looks like JIRA returns pretty verbose JSON...

deploystream/lib/transforms.py
txels: Maybe the middle ground is to add a parameter `keep_extra=True` if you want to keep the extra fields that JIRA may invoke with `False`?

tests/data/jira_issues.json
```diff
@@ -0,0 +1,360 @@
...
+                        "24x24": "https://something.atlassian.net/secure/useravatar?size=small&ownerId=alex.couper&avatarId=10603",
+                        "192x192": "https://something.atlassian.net/secure/useravatar?size=xxlarge%402x&ownerId=alex.couper&avatarId=10603",
+                        "16x16": "https://something.atlassian.net/secure/useravatar?size=xsmall&ownerId=alex.couper&avatarId=10603",
+                        "96x96": "https://something.atlassian.net/secure/useravatar?size=xxlarge&ownerId=alex.couper&avatarId=10603",
+                        "256x256": "https://something.atlassian.net/secure/useravatar?size=xxxlarge%402x&ownerId=alex.couper&avatarId=10603",
+                        "64x64": "https://something.atlassian.net/secure/useravatar?size=xlarge&ownerId=alex.couper&avatarId=10603",
+                        "128x128": "https://something.atlassian.net/secure/useravatar?size=xxxlarge&ownerId=alex.couper&avatarId=10603",
+                        "32x32": "https://something.atlassian.net/secure/useravatar?size=medium&ownerId=alex.couper&avatarId=10603"
```
txels, line 27: OMG what a waste of bandwidth

alexcouper:
_cough cough_

txels:
Hi, sorry I've had a shitty week. There is one outstanding comment about the semantic change in remapping. Have you tested that the github provider still works fine? I didn't really have time to test sprintly, may give it a go later.

alexcouper:
Ah yes, you wanted the option to `keep_extra`. I'll add that next time I'm working in this space

alexcouper:
That's sorted now.

Issue #71, opened 2013-05-01
Ability to compare two levels (to replace /deployable)


alexcouper:
List all branches & tags whose HEAD is in one tree but not in the other.

PR #73, opened 2013-05-07
Story/70/alex

alexcouper:
@txels This is _far_ from complete, but wanted to let you see it. Any comments welcome.

This is the beginning of solving #70 and #71 

Still needed at the back end to complete this:
- ~~An integration test to ensure we're getting the right hierarchies created.~~ See #77 
- ~~Writing of the `create_hierarchy_trees` (with a possible rename) to calculate which branches from each level are actually merged into higher ones and vice versa. This might end up needing to be moved to within a provider - or at least it will require extensive access to one.~~ See #82

Front end work:
- Extra design work on the front end to have this information displayed in an appropriate way.
- Addition of an angular route to see a feature and its details
- Altering of the feature list to incude urls to the deploystream version of the feature.

### DIFF
--- a/deploystream/apps/feature/lib.py
+++ b/deploystream/apps/feature/lib.py
@@ -40,13 +40,11 @@ def get_feature_info(feature_provider, feature_id, providers):
     :raises:
         UnknownProviderException - if no such name found.
     """
-    # First get feature info from the management provider
-
-    feature = None
-    try:
-        planning_provider = providers['by_name'][feature_provider]
-    except KeyError:
+    if feature_provider not in providers:
         raise UnknownProviderException(feature_provider)
+
+    # First get feature info from the management provider
+    planning_provider = providers[feature_provider]
 
     feature = Feature(planning_provider,
                       **planning_provider.get_feature_info(feature_id))
--- a/deploystream/apps/feature/lib.py
+++ b/deploystream/apps/feature/lib.py
@@ -22,13 +22,19 @@ def get_all_features(providers):
     return all_features
 
 
-def get_feature_info(feature_id, providers):
+def get_feature_info(feature_provider, feature_id, providers):
     """
     Get the information associated with the given feature from the providers
     given.
 
-    ``planning``, ``source_code`` and ``build_info`` are relevant providers to
-    be called.
+    :param feature_provider:
+        The name of the planning provider who knows of this feature.
+
+    :param feature_id:
+        The planning-provider specific id for the feature.
+
+    :param providers:
+        A dictionary of all providers.
     """
     # TODO: since features may come from various origins, we need
     # at this stage to either use a feature id that is a string such as
--- a/deploystream/apps/feature/lib.py
+++ b/deploystream/apps/feature/lib.py
@@ -39,7 +39,6 @@ def get_feature_info(feature_provider, feature_id, providers):
 
     :raises:
         UnknownProviderException - if no such name found.
-
     """
     # First get feature info from the management provider
 
@@ -54,12 +53,13 @@ def get_feature_info(feature_provider, feature_id, providers):
 
     # Then get any branch info from any source control providers
     for provider in providers[ISourceCodeControlProvider]:
-        for branch_data in provider.get_repo_branches_involved(feature_id):
+        for branch_data in provider.get_repo_branches_involved(
+            feature_id, app.config['HIERARCHY_REGEXES']):
             feature.add_branch(Branch(*branch_data, provider=provider))
 
     # Use that branch info, along with configuration regexes to create a
     # hierarchy of the branches involved in the feature.
-    feature.create_hierarchy_trees(app.config['HIERARCHY_REGEXES'])
+    feature.create_hierarchy_trees()
 
     # Ask source control providers for merging information at this point.
     for provider in providers[ISourceCodeControlProvider]:
--- a/deploystream/apps/feature/lib.py
+++ b/deploystream/apps/feature/lib.py
@@ -2,6 +2,7 @@
 #-*- coding: utf-8 -*-
 
 from deploystream import app
+from deploystream.exceptions import UnknownProviderException
 from deploystream.providers.interfaces import (
     IBuildInfoProvider, IPlanningProvider, ISourceCodeControlProvider)
 from .models import Branch, BuildInfo, Feature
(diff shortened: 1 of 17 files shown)


deploystream/apps/feature/lib.py
```diff
@@ -22,42 +23,43 @@ def get_all_features(providers):
...
+    # First get feature info from the management provider
+
     feature = None
-    for provider in providers[IPlanningProvider]:
-        feature = Feature(provider, None,
-                          **provider.get_feature_info(feature_id))
-        if feature:
-            break
+    try:
+        planning_provider = providers['by_name'][feature_provider]
```
txels: This may sound weird to you, but why don't we use a single dict that can have different types of keys, either strings when searching for provider name, or classes when searching for provider type. This is Python after all so we are not subject to single-type keys.

I see no danger in overloading the providers dictionary this way, and the code reads clearer.

deploystream/apps/feature/models.py
```diff
@@ -60,9 +60,12 @@ class Branch(object):
         ``branch_name``   - The name of the branch.
         ``latest_commit`` - The head commmit, or latest revision in this
                             branch.
+        ``level``         - The level this Branch falls in the hierarchy for
+                            the feature.
```
txels: clarify... is this a numeric depth level? Starts at 0?

deploystream/apps/feature/views.py
```diff
@@ -29,9 +30,13 @@ def list_features(providers):
     return features
 
 
-@app.route('/features/<feature_id>', methods=['GET'])
+@app.route('/features/<provider_feature_id>', methods=['GET'])
```
txels: Or `<provider>/<feature_id>` ... provider here refers more to origin (the planning provider where we learnt about this feature). 

But as we may get feature info from various providers of other types, I see this naming can be a source of confusion/lack of clarity in the future, once we eventually retrieve feature info specific to a provider (e.g. Jenkins job data for a feature).

deploystream/lib/hierarchy.py
```diff
@@ -0,0 +1,52 @@
...
+
+    :param hierarchical_regexes:
+        A list of regexes to be joined into one single regex.
+
+    :returns:
+        A single regex for easier matching.
+    """
+    subs = []
+    for index, regex in enumerate(hierarchical_regexes):
+        subs.append("(?P<level_{0}>{1})".format(index, regex))
```
txels: interesting approach

deploystream/lib/hierarchy.py
```diff
@@ -0,0 +1,52 @@
...
+        The feature to filter the branch names on.
+
+    :param branches:
+        A list of branch names to filter.
+
+    :param hierarchical_regexes:
+        A list of regexes assumed to be in descending order of branch status.
+
+    :returns:
+        The positional index that the branch should be found it. Or None if it
```
txels: `found in`?

deploystream/providers/github/__init__.py
```diff
@@ -85,3 +86,21 @@ def get_oauth_data(self):
                 'scope': 'repo'
             },
         }
+
+    def get_repo_branches_involved(self, feature_id, hierarchy_regexes):
+        branch_list = []
+
+        for repo in self.repositories:
+            for branch in repo.iter_branches():
```
txels, line 112: Interesting not to need git clones for this.
I've been thinking of the difficulties of eventually combining git/github providers. Won't be easy to come up with a cleanly layered solution.

deploystream/apps/feature/views.py
```diff
@@ -29,9 +30,13 @@ def list_features(providers):
...
 @needs_providers
 @as_json
-def view_feature(feature_id, providers):
-    feature = get_feature_info(feature_id, providers)
+def view_feature(provider_feature_id, providers):
+    provider_id, feature_id = provider_feature_id.split('...')
```
txels: Strange choice for a separator. Would single `'.'` not work? What about `:` or `;`?

deploystream/providers/github/__init__.py
txels, line 112: (or even just matching Jira/Sprintly projects to actual collections of repos in e.g. Github).

deploystream/providers/__init__.py
```diff
@@ -33,6 +33,7 @@ def get_providers(configs, session):
         interface
     """
     providers = defaultdict(list)
+    providers['by_name'] = {}
```
txels: just stick'em by name, a flat dict.

txels:
Looking good. If you wish we can split the work in smaller tasks, as e.g. just the front-end feature detail view seems like quite some work. I would like this to look much nicer than storyboard, so it would be nice to try to sketch out beforehand what we want to display and how (as opposed to getting a bunch of stuff in a template and try to make it look decent as an afterthought).

deploystream/apps/feature/lib.py
alexcouper: True - good idea.

deploystream/apps/feature/views.py
alexcouper: Are you saying that the variable name would be a source of confusion to developers, or the url structure to end users?

I'll change to a `/features/<source>/<feature_id>` url structure.

deploystream/apps/feature/views.py
alexcouper: I was obviously being inspired by git at the time - this will change as soon as I make the previous change.

deploystream/lib/hierarchy.py
alexcouper: interesting is an ambiguous word - but I'll assume that you're ok with it as is for now...

deploystream/providers/github/__init__.py
alexcouper, line 112: As long as we treat github repos as an independent source from the local git we should be fine. (Where independence means they have their own hierarchy trees etc)

alexcouper:
I think we should absolutely split the work into smaller tasks. I've updated the list of outstanding things in this PR and split them into front end / back end. By all means take the front end ones.
This is the point where github issues begin to show the failings - as it would be ideal to create sub-issues for each of them. Not sure how we want to manage them really.

One way would be to leave this PR open and if you're happy with it simply comment that you are and we then create PRs into this branch for the remaining work and we can therefore see when we're finished and be reminded of the outstanding subtasks...

deploystream/apps/feature/views.py
txels: Was thinking developers and variable names yes.

deploystream/lib/hierarchy.py
txels: it was good in this case, something I wouldn't have thought of

txels:
Good plan.

deploystream/providers/github/__init__.py
```diff
@@ -85,3 +86,21 @@ def get_oauth_data(self):
...
+                level = hierarchy.match_with_levels(
```
alexcouper, line 113: What do you think about this? It goes against the philosophy that I've been adopting to keep providers as dumb as possible.

An alternative would be that the provider defines a `def iter_branches(self)` method - returning repo name, branch name and head commit - and in https://github.com/pretenders/deploystream/pull/73/files#L0R54 we loop through each performing the relevant `hierarchy.match_with_levels` function.

The second approach would make test writing a lot easier too.

Which way do you prefer?

alexcouper, line 113: I think I'm going to take the view that each source code provider is responsible for it's own level producing - and that they may have different ways of calculating that - so therefore leaving the call to `match_with_levels` inside github's provider.

txels, line 113: OK I haven't given this enough thought TBH.

PR #74, opened 2013-05-07
Fix multi-project support in Sprint.ly

txels:
Now that we started tracking work in 2 products in sprintly, I realised there was a bug in the way features were retrieved from multiple projects.


deploystream/providers/sprintly/__init__.py
```diff
@@ -12,6 +12,7 @@
 FEATURE_MAP = {
     'number': 'id',
     'short_url': 'url',
+    ('product', 'name'): 'project',
```
alexcouper, line 15: Nice that this is needed for other people than JIRA's huge JSON dict.

PR #77, opened 2013-05-13
Dev/70/alex

alexcouper:
Add tests for getting github branches for a feature

### DIFF
--- a/deploystream/providers/github/__init__.py
+++ b/deploystream/providers/github/__init__.py
@@ -26,25 +26,39 @@ class GithubProvider(object):
     name = 'github'
     oauth_token_name = name
 
-    def __init__(self, token, organization=None, **kwargs):
+    def __init__(self, token, organization=None, repositories=None, **kwargs):
         """
         Initialise the provider by giving it GitHub credentials and repos.
 
         :param organization:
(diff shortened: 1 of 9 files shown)


deploystream/providers/github/__init__.py
```diff
@@ -26,21 +26,39 @@ class GithubProvider(object):
...
+
+        if organization:
             org = self.github.organization(organization)
             self.repositories = list(org.iter_repos())
+        elif repositories:
+            self.repositories = []
+            for owner, repo in repositories:
+                self.repositories.append(self.github.repository(owner, repo))
+        else:
+            self.repositories = list(self.github.iter_repos())
```
txels, line 61: Oh I just realised I did almost the same in PR #79 but worse. Will remove this bit from there.

txels, line 61: Oh no, there is a "subtle" difference, which is that I use the repo list to sort and filter repos even if organisation is not `None`. Which now that I think of it could be obtained by unsetting organization.

txels:
Happy to merge, although I have not looked at the whole branch matching thoroughly. But would like to see this globally merged to the "story" branch.

Issue #72, opened 2013-05-03
Speed issues


alexcouper:
I want this tool to be slick and quick.

It's starting to look nice, but it's slow to load. There are ways we can make it quicker.
### Break up the response

We could have the angular front end ask not for "all features" but instead "all features from this provider", followed by "all features from another" etc - so that the page appears in chunks. We'd still provide the option of asking for "all features" but just choose not to use it on our front end.
### Pre fetch and store.

I think when it comes to viewing a feature and calculating differences between branches, we're going to need to be doing things more cleverly on the server.

One way would be to ask for more permission from the github user (for example) to auto set up a callback on commit hook. We can then recalculate every time we hear about a commit for a user, and the data will be sitting there when they hit the site.

We should also break up the feature view and have any planning info brought back separately to the source code stuff.

PR #78, opened 2013-05-13
Add some level of caching for speed-up

txels:
I have added some caching to HTTP client calls (that use requests, by introducing requests-cache https://requests-cache.readthedocs.org/) and to views.

This is not really the goal of #72 (which is mostly about speeding up user feedback by progressive render) but is going to be quite useful in an active multi-user environment.

### DIFF
--- a/deploystream/lib/cache.py
+++ b/deploystream/lib/cache.py
@@ -1,0 +1,39 @@
+from functools import wraps
+
+from flask import request
+from werkzeug.contrib.cache import SimpleCache
+
+
+cache = SimpleCache()
+EXPIRY_SECONDS = 5 * 60
+
+
+def cached(timeout=EXPIRY_SECONDS, key='view/%s'):
+    """
+    A decorator for caching views.
+
+    Source: http://flask.pocoo.org/docs/patterns/viewdecorators/
+
+    With a little additional work it could be made into a generic caching
+    decorator for other functions that return pickleable data.
+    """
+    def decorator(f):
+        @wraps(f)
+        def decorated_function(*args, **kwargs):
+            cache_key = key % request.path
+            rv = cache.get(cache_key)
+            if rv is not None:
+                return rv
+            rv = f(*args, **kwargs)
+            cache.set(cache_key, rv, timeout=timeout)
+            return rv
+        return decorated_function
+    return decorator
+
+
+def activate_requests_caching():
+    """
+    Call once to activate caching for all HTTP GET issued by 'requests'
+    """
+    import requests_cache
+    requests_cache.install_cache(expire_after=EXPIRY_SECONDS)
--- a/deploystream/__init__.py
+++ b/deploystream/__init__.py
@@ -4,6 +4,7 @@ from os import environ
 from os.path import join, dirname
 from flask import Flask
 
+
 APP_DIR = dirname(__file__)
 CONFIG_DIR = join(dirname(APP_DIR), 'config')
 STATIC_DIR = join(APP_DIR, 'static')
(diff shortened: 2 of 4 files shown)


deploystream/lib/cache.py
```diff
@@ -0,0 +1,39 @@
...
+            if rv is not None:
+                return rv
+            rv = f(*args, **kwargs)
+            cache.set(cache_key, rv, timeout=timeout)
+            return rv
+        return decorated_function
+    return decorator
+
+
+def activate_requests_caching():
```
alexcouper, line 34: Why do we want both `activate_requests_caching` and `cached`?

Doesn't the `cached` decorator negate the need for the other?

txels, line 34: They serve different purposes. One caches what we serve to the user, the other caches what we get from other tools.
I believe there will be cases where what we deliver to the user will change (different views of the same data set) but what we need to fetch from the tools will be the same. Especially depending on how we get data from those tools (e.g. less requests that fetch a large dataset rather than many small requests - I think you were planning to use this kind of approach with fetching branch info from github).

User -(cached)- Ployst -(requests_caching)- 3rd party tools

Eventually we may want to review how we fetch and cache 3rd party data so that we can have some of it pre-fetched in the background, not impacting response time to the user. That lives in a longer-term future.

deploystream/lib/cache.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Source: http://flask.pocoo.org/docs/patterns/viewdecorators/
+
+    With a little additional work it could be made into a generic caching
+    decorator for other functions that return pickleable data.
+    """
+    def decorator(f):
+        @wraps(f)
+        def decorated_function(*args, **kwargs):
+            cache_key = key % request.path
```
alexcouper, line 23: I think this also wants to take into account user. Is there something in the request that is unique for the user?

deploystream/lib/cache.py
alexcouper, line 34: Yeah, I understood the difference in when they were caching - but in the use case of `list_features` the `activate_requests_caching` won't ever be used.

I'm a little uneasy with the way the `activate_requests_caching` does the caching - if it means all installed libraries using requests also have that cache then that might not be the right way about it. It feels like caching should be a per-provider thing - as source code info is more important to be live than planning info for example.

Happy to merge if you add a config param to decide whether to call this on line 70 of `deploystream/__init__.py` :)

txels, line 34: On retrospect, this looked like a quick win but it is a shoddy solution. I am concerned about the load time for the page and I want my team to use ployst and I didn't want page load time to be a pain point.

But I am not sure what is the best place to implement caching for provider calls. Either we do it by 1) wrapping it around calls to providers (preferrable, but at that level we may not be aware of what variables make output change - such as provider configuration) or 2) we leave that responsibility to providers' implementations (that will be more aware of what caching makes sense for the service they call, or which are the variables). 

Leaning towards 1), but changes in provider configuration may require busting the cache for that provider.

deploystream/lib/cache.py
txels, line 23: We will have to get session ID from somewhere?

txels:
At times like these I find myself missing the richness of Django as a framework and its extensive support for caching, e.g. its support for cache headers for browser caching could be also useful. I will have to do more research about proper ways to deal with that in Flask.

alexcouper:
It's not too late for us to change - (although it might be better to spend
our time elsewhere at the moment).

Where possible we're going to have to pre-fetch to achieve a decent user
experience. Github is easily achievable in that way. The planning tools are
slightly trickier as I doubt any provide a hook url to post to when changes
are made - but they have the added bonus of not often changing (until we
want to display task status etc) so maybe we can get away with moderately
frequent pre-fetches of that stuff?

txels:
Oh no I didn't mean we should change to Django at all. Just that some things take longer to figure out. Happy about this project's learning curve so far.

PR #83, opened 2013-05-25
Dev/70/carles show details

txels:
This is not intended to be merged, but mainly for discussion. You can also use it to add some crude visuals to the API.

When trying to glue things together in a quick first initial story detail view, I realised the JSON data structure we are generating right now (nested dictionaries) is not ideally suited for iterating in Javascript and Angular-JS templates. See my inline comments in the PR. 

We may want to come up with a properly defined and documented feature JSON data structure (and in particular all attachment info: repos, branches, build results...) before going much further.

### DIFF
--- a/config/local_settings_pretenders.py
+++ b/config/local_settings_pretenders.py
@@ -1,0 +1,20 @@
+# Settings for pretenders projects
+
+GITHUB_CONFIG = {
+    'organization': 'pretenders',
+}
+
+# GIT_CONFIG = {
+#     'feature_breakup_regex': "(?P<project>[a-zA-Z]+)-?(?P<id>[0-9]+)",
+#     'branch_finder_template': ".*(?i){project}.*"
+# }
+
+HIERARCHY_REGEXES = [
+    'master',
+    'develop',
+    'story/{FEATURE_ID}',
+    'story/{FEATURE_ID}/[a-z_]*',
+    '{PARENT}[/_][a-z]*',
+    'dev/{FEATURE_ID}/[a-z]*',
+    '{PARENT}[/_][a-z]*',
+]
--- a/deploystream/apps/feature/models.py
+++ b/deploystream/apps/feature/models.py
@@ -42,11 +42,13 @@ class Feature(object):
         self._extras = kwargs
 
         self.branches = defaultdict(dict)
+        self.branch_list = []
         self.trees = []
 
     def add_branch(self, branch):
         assert isinstance(branch, Branch)
         self.branches[branch.repo_name][branch.branch_name] = branch
+        self.branch_list.append(branch)
 
     def create_hierarchy_trees(self):
         "Create hierarchy trees - one for each repo."
--- a/deploystream/apps/feature/views.py
+++ b/deploystream/apps/feature/views.py
@@ -30,7 +30,7 @@ def list_features(providers):
     return features
 
 
-@app.route('/features/<source_id>/<feature_id>', methods=['GET'])
+@app.route('/features/<source_id>-<feature_id>', methods=['GET'])
 @needs_providers
 @as_json
 def view_feature(source_id, feature_id, providers):
--- a/deploystream/static/github-70.json
+++ b/deploystream/static/github-70.json
@@ -32,12 +32,17 @@
             "health": {
                 "status": "green",
                 "details": [
-                    {
+                  {
                     "type": "build",
                     "provider": "travis-ci",
                     "status": "green",
                     "url": "https://travis-ci.org/pretenders/deploystream/builds/8604410"
-                    }
+                  },
+                  {
+                    "type": "relevant",
+                    "provider": "git",
+                    "status": "green"
+                  }
                 ]
             },
             "children": [
@@ -52,6 +57,17 @@
                       "provider": "travis-ci",
                       "status": "red",
                       "url": "https://travis-ci.org/pretenders/deploystream/builds/8604410"
+                    },
+                    {
+                      "type": "relevant",
+                      "provider": "git",
+                      "status": "green"
+                    },
+                    {
+                      "type": "updated",
+                      "provider": "git",
+                      "status": "yellow",
+                      "message": "this branch is not up to date with develop"
                     }
                   ]
                 },
--- a/deploystream/static/github-70.json
+++ b/deploystream/static/github-70.json
@@ -3,7 +3,7 @@
   "title": "Ability to view the branch hierarchy for a feature, and what is merged into what.",
   "url": "https://github.com/pretenders/deploystream/issues/70",
   "project": "pretenders/deploystream",
-  "owner": "",
+  "owner": "alexcouper",
   "provider": "github",
   "type": "story",
   "id": 70,
--- a/deploystream/static/github-70.json
+++ b/deploystream/static/github-70.json
@@ -43,7 +43,7 @@
             "children": [
                 {
                 "latest_commit": "98457ab",
-                "branch_name": "story/70/alex",
+                "branch_name": "story/70/view-hierarchy",
                 "health": {
                   "status": "red",
                   "details": [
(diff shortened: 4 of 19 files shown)


config/local_settings_pretenders.py
```diff
@@ -0,0 +1,20 @@
+# Settings for pretenders projects
```
txels, line 1: Instead of using our personal ones, I believe this should be the reference settings file that we use when working on code improvements for everything github/git related, which I believe is going to be most of our next body of work.

deploystream/apps/feature/models.py
```diff
@@ -42,11 +42,13 @@ def __init__(self, provider, project, id, title,
         self._extras = kwargs
 
         self.branches = defaultdict(dict)
+        self.branch_list = []
         self.trees = []
 
     def add_branch(self, branch):
         assert isinstance(branch, Branch)
         self.branches[branch.repo_name][branch.branch_name] = branch
```
txels, line 50: This data structure doesn't lend itself well to being iterated in Javascript / front-end code. For want of something better, I am using a flat list of branches for now (see line below).

deploystream/apps/feature/views.py
```diff
@@ -30,7 +30,7 @@ def list_features(providers):
     return features
 
 
-@app.route('/features/<source_id>/<feature_id>', methods=['GET'])
+@app.route('/features/<source_id>-<feature_id>', methods=['GET'])
```
txels, line 33: The `/` was a bit of a pain when trying to set routes and use a single angular-js resource for `Feature`.

deploystream/static/js/controllers.js
```diff
@@ -11,4 +11,24 @@ function FeatureListCtrl($scope, Feature) {
...
+    Feature.query('', function (features) {
+        $scope.features = features;
+    });
+
+    Feature.get({featureId: featureId}, function (feature) {
+        var repos = [];
+        var branches = feature.branch_list;
+        $scope.feature = feature;
+        $scope.branches = branches;
+        $scope.root = 'master';
```
txels, line 26: This is something that will need to come from the backend, as part of the feature branch data.
As you have been taking great care to provide properly hierarchified branch info, the frontend code only needs the root branch, and work with its `children`. A lot of the branch data that comes in the JSON is redundant and I don't think we really need it.

deploystream/static/js/services.js
```diff
@@ -11,6 +11,7 @@ angular.module('deploystream.services', ['ngResource'])
     .factory('Feature', function ($resource) {
         return $resource(
             '/features/:featureId',
+            //'/static/:featureId.json',
```
txels: (I only use this occasionally for testing feature detail page rendering using a static JSON file, as else refresh takes too long).
I am thinking about using some pretenders mocking of backend, but it may be tricky as the frontend code is actually the same host and port. But they can be decoupled for testing.

deploystream/static/partials/feature.html
```diff
@@ -0,0 +1,41 @@
...
+    <p>
+        {{feature.project}}
+        <span class="label label-info">{{feature.type}}</span>
+    </p>
+    <p>
+        Owner: {{feature.owner}}
+    </p>
+
+    <h3>Branches</h3>
+    <script type="text/ng-template"  id="branch.html">
```
txels: this strange construct is actually a recursive template that I use to render the branch tree.

deploystream/static/partials/feature.html
```diff
@@ -0,0 +1,41 @@
...
+        <p>
+            {{branch.branch_name}}
+            <span class='label'>{{branch.latest_commit}}</span>
+        </p>
+        <ul>
+            <li ng-repeat="branch in branch.children" ng-include="'branch.html'"> </li>
```
txels, line 44: (recursion happens here)

deploystream/apps/feature/views.py
alexcouper, line 33: Interesting. I would have expected `/features/:sourceid/:featureId` to work

alexcouper:
Looks nice - I get this at the moment. I guess that's because nothing has been merged into anything else yet?

screen shot 2013-05-27 at 1 10 24 pm

deploystream/static/partials/feature.html
alexcouper: Could you put a wee comment into the template that doesn't get rendered (if angular allows that) saying this?

deploystream/static/js/controllers.js
alexcouper, line 26: Should we address this now?

alexcouper:
We also need to handle 404s when a feature is looked up that does not exist.

At the backend, I'm putting the logic in place to return a 404. Angular needs to handle this and display something appropriately (ATM it just continues to show the spinning wheel)

deploystream/apps/feature/views.py
txels, line 33: It's more related to using the `Feature` resource in angular to retrieve both detail and list (when `featureId` is empty). Also trying to build `":provider/:featureId"` in JS was resulting in a URL-escaped value for `/` that returned a 404.

deploystream/static/partials/feature.html
txels: Angular templates are HTML that gets served as-is, so it will have to be plain old HTML comments `<!-- -->`

deploystream/static/js/controllers.js
txels, line 26: Maybe. I believe at the feature "repo" level we may have a couple of pieces of data that will be useful: 1) which is the root branch of the hierarchy to display and 2) which is the "main" feature branch for this feature - as both have display relevance in different situations.

txels:
I just updated this branch with the mock-up of a more suitable JSON response, and some front-end that starts to resemble what I sketched out on paper

Run `./scripts/compile-css.sh` to update the CSS and point browser to http://localhost:5000/#/features/github-70

txels:
Forgot to mention, the focus of the discussion on this PR has shifted to getting consensus on:
- the JSON API output for a feature, here's the mock-up: https://github.com/pretenders/deploystream/pull/83/files#diff-5
- the visual output which should look as attached:

ployst

deploystream/static/github-70.json
```diff
@@ -0,0 +1,138 @@
+{
```
alexcouper, line 1: Structure seems fine to me.

Presumably additional details could be added to the `details` list. Not sure how we'd display that additional information since it seems we are only showing one overall colour but that's one for another day I guess...

deploystream/static/partials/feature.html
```diff
@@ -0,0 +1,42 @@
...
+    <p>
+        Owner: {{feature.owner}}
+    </p>
+
+    <h3>Branches</h3>
+    <script type="text/ng-template" id="branch.html">
+        <p>
+            <i class="icon icon-caret-right"></i>
+            <i class="icon icon-large icon-sign-blank health-{{branch.health.status}}"></i>
+            <i class="icon icon-wrench"></i>
```
alexcouper: Is the wrench specific to showing the status is related to a build?

alexcouper:
I think the visual side is a great start. There's a lot of white space going to be created as a result of the way the branching works, but I'm fine with that for now.

deploystream/static/github-70.json
txels, line 1: The main idea as described in the original google doc was: "Status/health is a single square with a color code. It represents an aggregated view of a number of indicators. Upon clicking/hovering one can see the detailed decomposition (e.g. build status, branch relevance, branch staleness, ready to deploy...) "

I will add some more details for illustration and completeness.

deploystream/static/partials/feature.html
txels: This is supposed to be a tool button that should unfold a menu of actions that can be done with this branch (such as trigger a [re]build, send PR to parent, deploy to some server, etc...). I can make it look more like a button if you think that's better.

txels:
I am sure we can try and find ways to make better use of screen, and reduce white space without generating clutter. There are a few blank ares... What white space are you specifically thinking about at this moment?

alexcouper:
The white space to the right of the set of branches. I think in time other information to do with the feature could go there, so it's not a biggie.

PR #82, opened 2013-05-24
Dev/70/alex hierarchy

alexcouper:
This works out the hierarchy of branches and exposes them correctly in the API.

`get_feature_info` needs to be written in the github provider.

PR #84, opened 2013-05-27
Dev/70/alex get feature info

alexcouper:
Look at this one after #82

### DIFF
--- a/deploystream/apps/feature/lib.py
+++ b/deploystream/apps/feature/lib.py
@@ -41,7 +41,6 @@ def get_feature_info(feature_provider, feature_id, providers):
     :raises:
         UnknownProviderException - if no such name found.
     """
-    print providers
     if feature_provider not in providers:
         raise UnknownProviderException(feature_provider)
 
@@ -49,7 +48,7 @@ def get_feature_info(feature_provider, feature_id, providers):
     planning_provider = providers[feature_provider]
 
     feature_info = planning_provider.get_feature_info(feature_id)
-    print planning_provider, feature_info
+
     if not feature_info:
         raise UnknownFeatureException(feature_id)
 
--- a/deploystream/apps/feature/lib.py
+++ b/deploystream/apps/feature/lib.py
@@ -2,7 +2,8 @@
 #-*- coding: utf-8 -*-
 
 from deploystream import app
-from deploystream.exceptions import UnknownProviderException
+from deploystream.exceptions import (
+    UnknownProviderException, UnknownFeatureException)
 from deploystream.providers.interfaces import (
     IBuildInfoProvider, IPlanningProvider, ISourceCodeControlProvider)
 from .models import Branch, BuildInfo, Feature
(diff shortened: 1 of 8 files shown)


deploystream/providers/github/__init__.py
```diff
@@ -106,19 +123,75 @@ def get_oauth_data(self):
...
+                        # we haven't already done so and store them in the
+                        # temporary ``repo_branches`` dict
+                        if repo_branches[sha].get('commits') is None:
+                            c_list = []
+                            for commit in repo.iter_commits(sha=sha):
+                                commit_date = commit.commit.committer['date']
+                                commit_date_time = iso8601.parse_date(
+                                    commit_date)
+                                if (commit_date_time.replace(tzinfo=None) <
+                                        two_months_ago):
```
alexcouper, line 179: For now I'm using this to stop going back to the beginning of time for every branch that matches the feature id.

Sooner rather than later we're going to need a chat about switching the architecture to being pre-fetched. Are you free at all this week?

txels:
Github is not figuring out the isolated changes after your other PR was merged, which makes the intent of this PR harder to follow.

alexcouper:
@txels that's better.

deploystream/apps/feature/lib.py
```diff
@@ -46,8 +47,13 @@ def get_feature_info(feature_provider, feature_id, providers):
...
     planning_provider = providers[feature_provider]
 
-    feature = Feature(planning_provider,
-                      **planning_provider.get_feature_info(feature_id))
+    feature_info = planning_provider.get_feature_info(feature_id)
+
+    if not feature_info:
+        raise UnknownFeatureException(feature_id)
+
+    feature = Feature(planning_provider, **feature_info)
```
txels, line 55: There is something still nagging me with this approach. 

When getting feature info from a provider this way, it will make it more difficult to correlate and aggregate data from various sources (e.g. planning info in sprintly, open PRs in github, branches in github/git, jobs in Jenkins...). An alternative approach is passing a `Feature` object (or a dict if you prefer) around to the various providers in sequence (starting with the planning provider) so that each _decorates_ the feature with additional data it knows about.

For example, I would like to get initial feature info from sprintly, then pass this to the github provider so that it adds information about open PRs and relevant branches (and even mergeability status and build success from the PR), then to Jenkins so that it can add information about known builds to the feature's branches...

Up for hangout discussion.

deploystream/providers/github/__init__.py
txels, line 179: I think the best local caching for repos is going to be local clones as in storyboard, which will make all these more efficient.

Give this a read and see possible solutions, I am thinking deploy keys but there are other options: https://help.github.com/articles/managing-deploy-keys

One can create deploy keys via the API: http://developer.github.com/v3/repos/keys/

deploystream/providers/github/__init__.py
```diff
@@ -90,13 +81,34 @@ def get_features(self, **filters):
...
-        # For now we'll do a really crude search through the get_features
-        # results
-        for feat in self.get_features():
-            if str(feat['id']) == str(feature_id):
-                return feat
+        # Issue with this approach is that we return the first issue with an
+        # ID across all repos.
+        # Such are the shortcomings of using Git as a planning provider. To
+        # get round this we'd need to have the repo in the feature_id, but this
+        # seems a bad idea from the POV of matching branch names.
```
txels, line 105: Yup this is a hairy issue. We may want to think of a decent solution to that.

deploystream/providers/github/__init__.py
alexcouper, line 179: We'll revisit this in the next round of changes following the hangout I guess.

I'm going to set up some planning tool - I think we could do with some stories & tasks to get this going.

PR #85, opened 2013-05-31
Beginnings of user accounts, sqlalchemy etc.

alexcouper:
To create the DB run the `create_db.py` script.

There's work to be done on getting the login, register and front page to all look good.

I was thinking that we could go for the github approach that assumes you'll mostly stay logged in. ie. the front page has a registration field on it, with a button at the top for signing in.

However, I think I'm missing a bunch of less files from bootstrap so wanted to talk to you about that - feel free on taking that task though!

Unfortunately asana.com doesn't provide decent issue numbers (they're super long like 5641387051775) so we may need to move away from that. Someone on twitter recommended binfire.com to me. Thoughts?

[I've added some failing tests so that I know what I need to finish up when I get back to this]

PR #87, opened 2013-06-01
Alternative login form as a popover

txels:
This is an alternative to #86 to style #85 

See if you like this better (I do). It uses a popover to show the login form, which is retrieved via Ajax.

If you like this I will clean it up a bit so that it is not a quick hack.


alexcouper:
This looks really good. :+1:

alexcouper:
I've merged this for now - you can clean it up in another PR...

PR #89, opened 2013-06-23
Dev/user signup/alex

alexcouper:
This moves the profile page into Angular land.

There's no tests at the angular end added as part of this :/

It's unclear to me how we want to structure the angular world stuff.
eg. the oauths stuff in controllers.js should be in another function, but where should that live?
and the app.js list of providers is gradually going to get unmanageable.

Do you think that registration and login should be moved inside the Angular app or be templates rendered through flask?

### DIFF
--- a/deploystream/apps/users/views.py
+++ b/deploystream/apps/users/views.py
@@ -1,15 +1,11 @@
 from flask import (Blueprint, request, render_template, flash, g, session,
     redirect, url_for)
-from werkzeug import check_password_hash, generate_password_hash
-
-from deploystream import db
-from deploystream.lib.transforms import as_json
+from werkzeug import check_password_hash
 
 from .forms import RegisterForm, LoginForm
 from .models import User
 from .lib import load_user_to_session
 from .decorators import requires_login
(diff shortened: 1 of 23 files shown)


alexcouper:
I'm going to start working on repo configuration now, so have created a story branch from this branch.

deploystream/apps/users/views.py
```diff
@@ -28,6 +21,12 @@ def before_request():
         g.user = User.query.get(session['user_id'])
 
 
+@mod.route('/me/')
+@requires_login
+def home():
```
txels, line 26: This is now a bit weird as this URL was initially an "HTML" view, and now is a "JSON" view - this is where users are redirected after login.

Login redirection should now be `/` or `/#/profile`, I guess... ?

Is there any reason for this URL to exist at all, though?

alexcouper, line 26: I was thinking it would be `/#/profile`. This PR includes that change I believe.

txels:
All good, see PR #91 for the case I meant

PR #92, opened 2013-06-30
Fix error in Oauth user creation return value

txels:
I just found this bug in a situation where I could no longer "login via github". Additionally, I was getting a missing token exception when listing features.
- `get_or_create_user_oauth` was returning a user_id while the caller expected a user instance.

This would need a regression test, but I did not have time to write it.

An additional problem left is the fact that we now generate Oauth 302 redirects as a result of API calls, which are not handled by the frontend. I guess that with the new way of handling user accounts and tokens, Oauth redirects should not happen at the points where we are issuing API calls, only at the point of generating tokens when assimilating a github account.


alexcouper:
Good catch. Merging,

PR #95, opened 2013-07-31
Change User<>Repo to be 1-Many. Add Repo creation ability. Add tests.

alexcouper:
I decided to change the User-Repo relationship on the basis that the many-to-many was a pre-optimization step. 

If the time comes that this needs to be changed then so be it.

### DIFF
--- a/deploystream/apps/repo/api.py
+++ b/deploystream/apps/repo/api.py
@@ -2,9 +2,17 @@ from flask import session
 from flask.ext.restless import APIManager, ProcessingException
 from sqlalchemy.orm.exc import NoResultFound
 from deploystream import app, db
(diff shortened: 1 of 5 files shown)


txels:
Yes totally agree

txels:
Looks sane to me, although I have only skimmed through the code. Will take a decent look later this week.

txels:
Travis is lying. We need to improve our test script.

txels:
I did a quick fix to the test script so that Travis actually fails when tests fail.

alexcouper:
Nice. I'll need to fix the version of flask to pre the removal of exceptions until flask restless releases an update.

The good news is that the fix in restless has been made so it wont be too long.

